PR #23, opened 2014-01-19
add optional wigle OSM tile source, on by default

uhtu:
choice of evils

### DIFF
--- a/res/layout/settings.xml
+++ b/res/layout/settings.xml
@@ -96,6 +96,11 @@
         android:layout_height="wrap_content"
         android:text="@string/use_network_location" />
 
+<CheckBox android:id="@+id/use_wigle_tiles"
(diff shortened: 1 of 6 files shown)


slammingprogramming:
I notice that the FOSS branch is placed into a lower usefulness due to the lack of API integration with a mapping platform, as the main branch relies on Google Maps, the FOSS version has none implemented at all, and as such a essential and core function of the app, viewing the data overlaid on a map, has been eliminated from the FOSS version, leaving one to choose between FOSS but missing features, or mainline but using Google services. I noticed this PR during my search for this being already considered and rejected, but was surprised to find it is actually one that was accepted in 2014. I just would like to ask whether this was removed from the FOSS version for a reason and whether its reimplementation would be acceptable to the community and maintainers.

PR #64, opened 2016-02-20
Consistently use “WiFi”

Aeyoun:
Change instances of “Wifi” to “”WiFi”. Makes the app consistent
with the Android system and with itself.

Change hyphens to en-dashes in titles.


bobzilladev:
sure!

PR #73, opened 2016-07-26
Updated German description translation

kill0rz:
Updated the translation of the description file.

Seems like this was previously translated automatically. Horrible grammar. :smiley: 
Hope that this is better now.


bobzilladev:
Aye, as noted in this file, it's all machine-translated unless a kind user comes along and makes it better :)

kill0rz:
Very glad you accept corrections. ;)

PR #74, opened 2016-07-28
added some missing German language variables

kill0rz:


bobzilladev:
Nice, grabbed all the new text too. Thanks!

PR #83, opened 2017-01-03
FEA - cutting news over to new API

rksh:
Required organizing GET/POST method partition, since all previous requests were performed with POST. Note that this required significant partitioning of connection configuration in HttpFileUploader as well as flowing method throughout the system.

Please review, include critical feedback. Notably, I believe that HttpFileUploader/HttpDownloader were somewhat mis-named even before this PR.


bobzilladev:
Seems legit as long as it tests happily. I'd pry make GET/POST a constant or better yet an enum, or borrow an enum if one is in classpath somewhere (I assume there's something somewhere already).

bobzilladev:
+1

PR #84, opened 2017-01-05
wigle

undertow666:


bobzilladev:
Hi! If you want to take this on, you'll need to resolve the conflicts (the hard part of maintaining any fork). The apk can be ignored since a new one will be generated for this branch. The other files may have obvious ways to detangle, since basically the only thing this foss-master branch does is remove references to google maps to help out the foss purists. Then the source needs to be built and tested.

bobzilladev:
I'm going to close this for now since it has conflicts and more work to take care of. If you do want to head down this path we would really appreciate all the effort!

PR #85, opened 2017-01-13
BUG - reset tokens on credentials change

rksh:
re-authenticate if a user changes username or password.
this is accomplished by clearing cached auth user/token and clearing cached files

### DIFF
--- a/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/NewsFragment.java
+++ b/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/NewsFragment.java
@@ -98,7 +98,10 @@ public class NewsFragment extends Fragment {
                 });
         try {
             task.startDownload(this);
-        } catch (WiGLEAuthException waex) {} //unauthenticated call
+        } catch (WiGLEAuthException waex) {
+            //unauthenticated call - should never trip
+            MainActivity.warn("Authentication error on news load (should not happen)", waex);
+        }
 
         return rootView;
     }
--- a/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/NewsFragment.java
+++ b/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/NewsFragment.java
@@ -96,7 +96,9 @@ public class NewsFragment extends Fragment {
                         handleNews(json, handler);
                     }
                 });
-        task.startDownload(this);
+        try {
+            task.startDownload(this);
+        } catch (WiGLEAuthException waex) {} //unauthenticated call
 
         return rootView;
     }
--- a/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/SettingsFragment.java
+++ b/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/SettingsFragment.java
@@ -1,13 +1,17 @@
 package net.wigle.wigleandroid;
 
+import java.io.File;
+import java.io.FilenameFilter;
 import java.util.Arrays;
 
 import android.annotation.SuppressLint;
+import android.app.ListActivity;
 import android.content.Context;
 import android.content.Intent;
 import android.content.SharedPreferences;
 import android.content.SharedPreferences.Editor;
 import android.media.AudioManager;
+import android.os.Build;
 import android.os.Bundle;
 import android.support.v4.app.Fragment;
 import android.text.Editable;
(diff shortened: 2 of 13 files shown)


wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/SettingsFragment.java
```diff
@@ -160,10 +164,17 @@ public void onCheckedChanged( final CompoundButton buttonView, final boolean isC
...
-        }
-        catch (Exception ex) {
+            if (Build.VERSION.SDK_INT >= 24) {
+                //Html.fromHtml(String, int) // for 24 api and more
+                register.setText(Html.fromHtml("<a href='https://wigle.net/register'>" + registerString + "</a> "
+                        + atString + " <a href='https://wigle.net/register'>WiGLE.net</a>",
+                        Html.FROM_HTML_MODE_LEGACY));
+            } else {
+                register.setText(Html.fromHtml("<a href='https://wigle.net/register'>" +
+                        registerString + "</a> " + atString +
```
bobzilladev: Looks like this big string is the same in both branches, can it be moved before the 'if' for less duplication?

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/background/ApiDownloader.java
```diff
@@ -231,18 +231,20 @@ private void downloadTokenAndStart(final Fragment fragment) {
                     public void requestComplete(final JSONObject json, final boolean isCache) {
                         try {
                             // {"success":true,"authname":"AID...","token":"..."}
+                            final SharedPreferences prefs = fragment.getContext()
+                                    .getSharedPreferences(ListFragment.SHARED_PREFS, 0);
```
bobzilladev: Doesn't look like this prefs needs to move, can pry stay close to where it is used. Guessing came here organically while you were playing with different solutions.

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/SettingsFragment.java
rksh: totes.

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/NewsFragment.java
```diff
@@ -96,7 +96,9 @@ public void requestComplete(final JSONObject json, final boolean isCache) {
                         handleNews(json, handler);
                     }
                 });
-        task.startDownload(this);
+        try {
+            task.startDownload(this);
+        } catch (WiGLEAuthException waex) {} //unauthenticated call
```
bobzilladev: Empty catches always wig me out, can we at least have an info log?

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/SettingsFragment.java
```diff
@@ -345,6 +363,23 @@ private void updateRegister(final View view) {
         }
     }
 
+    /**
+     * clear cache files (i.e. on creds change)
+     */
+    private void clearCachefiles() {
+        final File cacheDir = new File(MainActivity.getSDPath());
```
bobzilladev: Did you test this on both sd card (allowing access to photos, downloads, etc). And also not (when it uses the app-specific storage area).

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/SettingsFragment.java
```diff
@@ -200,7 +214,11 @@ public void onCheckedChanged( final CompoundButton buttonView, final boolean isC
             public void onTextChanged( final String s ) {
                 // ListActivity.debug("pass: " + s);
                 editor.putString( ListFragment.PREF_PASSWORD, s.trim() );
+                // ALIBI: if the password changes, refetch token
+                editor.remove(ListFragment.PREF_AUTHNAME);
+                editor.remove(ListFragment.PREF_TOKEN);
                 editor.apply();
+                clearCachefiles();
```
bobzilladev: Nit picky, but it comes up in my brain so I guess I'll say it. Both of these "clear stuff out" paths are the same, could consolidate.

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/SiteStatsFragment.java
```diff
@@ -120,7 +120,9 @@ public void requestComplete(final JSONObject json, final boolean isCache) {
                         handleSiteStats(json, handler);
                     }
                 });
-        task.startDownload(this);
+        try {
+            task.startDownload(this);
+        } catch (WiGLEAuthException waex) {} //unauthenticated call
```
bobzilladev: Info log just in case if you would.

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/SettingsFragment.java
rksh: hah, solid catch. in fact, the RO handling was ok, but the no-cache-files case was broken.

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/SettingsFragment.java
rksh: only three lines are actually duplicated. generalizing the prefs-clear + a new method (parameterize field, new value and prefs) makes DRYing a dead heat.

bobzilladev: I don't care about lines, but duplicated logic. When something else is found that needs to be cleared, will that person update both copies of the logic? maybe...

PR #86, opened 2017-01-19
ENH - color and layout tweaks

rksh:
Making the stats and upload views match site color schemes, tweaking layout.


bobzilladev:
I assume that'll look delightful 👍

PR #88, opened 2017-01-25
ENH - A first cut at camera-based client activation using a logged-in qr on WiGLE site

rksh:
question: is there a way to make this work (but exclude the activation link) if the play APIs aren't available on build/device?


wiglewifiwardriving/src/main/res/values/strings.xml
```diff
@@ -319,5 +317,7 @@
     <string name="prevrank_title">User Ranking Last Month</string>
     <string name="prevmonthrank_title">Month Ranking Last Month</string>
     <string name="list_scanning_on">(Scanning turned on)</string>
-
+    <string name="activate">activate now</string>
+    <string name="title_activity_activate">ActivateActivity</string>
```
bobzilladev: Can you put this title_activity_activate last, since it isn't being added to the other strings files? and now there is a gap in some of the other strings files instead of just truncation, which might be worse. would pry want to copy in the english version to those ones?

bobzilladev:
if play isn't on the device we pop up a dialog asking to have it installed. unfortunate that this requires 2 more permissions, on older platforms that'll require folks to approve the update which we always lose people over. but i spose they aren't that active anyway. progress! this is cool stuff.

rksh:
Since the API only exists in 4.2.2-and-up, is there a way to make the permissions conditional (since we can't use the feature on platforms < SDK = 17  anyway?)

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/SettingsFragment.java
```diff
@@ -161,9 +247,15 @@ public void onCheckedChanged( final CompoundButton buttonView, final boolean isC
...
         final String registerString = getString(R.string.register);
-        final String atString = getString(R.string.at);
-        final String registerBlurb = "<a href='https://wigle.net/register'>" + registerString +
-                "</a> " + atString + " <a href='https://wigle.net/register'>WiGLE.net</a>";
+        final String activateString = getString(R.string.activate);
+        String registerBlurb = "<a href='https://wigle.net/register'>" + registerString +
+                "</a> @WiGLE.net";
+
+        // ALIBI: vision APIs started in 4.2.2; JB2 4.3 = 18 is safe. 17 might work. -rksh
+        if (Build.VERSION.SDK_INT >= 18) {
```
rksh: if `uses-permission-sdk-23` in AndroidManifest.xml makes sense, this should be updated accordingly.

bobzilladev:
I'm all for this with the uses-permission-sdk-23 magick.

rksh:
will be merge-able once https://github.com/wiglenet/wigle-wifi-wardriving/pull/91 is complete and incorporated.

rksh:
SHIP IT

PR #89, opened 2017-01-29
ENH - cutting over the API-based observation retrieval

rksh:
surprisingly complex. some efforts made to bring together ApiDownloader and ObservationImporter (former HttpDownloader) - however some differences still outstanding


bobzilladev:
This is getting pretty big, but all looks happy. There's some repeating, but it's noted in comments as future dry'ing. lgtm.

PR #90, opened 2017-02-01
Fixes #78

### LINKED ISSUE #78: HTML and linebreak translation in news view doesn't work.
In news view, HTML and linebreaks (\n) aren't translated so these codes are shown as plain text.

pic1
pic1

rksh:
wonky BBcode rewriter, using from-html conversion.
fixes linebreak translation (and other inconsistent entities) per https://github.com/wiglenet/wigle-wifi-wardriving/issues/78


bobzilladev:
wow, nice. +1

rksh:
woot!

PR #91, opened 2017-02-07
ENH - upload using the new API

rksh:
in order to switch over to QR-based activation, we need everything to use the same API pathway. This is a "light-weight" cutover to the v2 API for uploads without undertaking a deeper rewrite to limit scope and complexity of the change. It's clear that we'll need to further refactor AbstractApiRequest children to bring the observation upload/download functionality into line with the ApiDownload functionality, and DRY everything up. This still requires further edge-case testing, and review. Steps to be taken after this change (not included in this PR):
- move static methods out of HttpFileUploader (and either into their own utility class or into AbstractApiRequest)
- Eliminate vestigial HttpFileUploader and FileUploaderTask


bobzilladev:
looks legit

rksh:
SHIP IT

PR #92, opened 2017-02-08
ENH - build script updates

rksh:
Here are some things my dev environment keeps re-including whenever i revert and start developing. they seem innocuous. feel free to reject.


rksh:
merging per discussion

PR #93, opened 2017-02-08
FIX - removing redundant logging

rksh:
these were bugging me.


rksh:
if this breaks things... well, there's that.

PR #95, opened 2017-02-11
ENH- i18n template values and user-comprehensible strings for upload status

rksh:
Making the uploads screen display something people can read for status instead of single-letter codes.

### DIFF
--- a/wiglewifiwardriving/src/main/res/values-cs/strings.xml
+++ b/wiglewifiwardriving/src/main/res/values-cs/strings.xml
@@ -257,9 +257,9 @@
     <string name="activate">aktivujte nyní</string>
     <string name="setting_activate_prompt">visit https://wigle.net/activate for your user code</string>
     <string name="upload_queued">Queued</string>
-    <string name="upload_parsing">Parsing...</string>
-    <string name="upload_trilaterating">Trilaterating...</string>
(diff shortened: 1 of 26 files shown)


wiglewifiwardriving/src/main/res/values-cs/strings.xml
```diff
@@ -256,5 +256,11 @@
     <string name="disable_toast">Disable toast pop-ups</string>
     <string name="activate">aktivujte nyní</string>
     <string name="setting_activate_prompt">visit https://wigle.net/activate for your user code</string>
-
+    <string name="upload_queued">Queued</string>
+    <string name="upload_parsing">Parsing...</string>
+    <string name="upload_trilaterating">Trilaterating...</string>
+    <string name="upload_stats">Final Stats...</string>
```
bobzilladev: could use the elipse single character, might make it fit a little better. oh hell, this is on a million files tho. i18n ftl.

bobzilladev:
krad, nice polish, +27b/6

wiglewifiwardriving/src/main/res/values-cs/strings.xml
rksh: done as suggested

bobzilladev:
I'm Lovin' It

PR #97, opened 2017-02-17
ENH - Allow update based on network strength

cognitive-rift:
This change allows the database to be updated based on signal strength as a primary condition, as opposed to a secondary condition to lat/lon changes.

This will cause more updates to occur, but the resulting data will be far more accurate.

Only possible hiccup I could imagine is the LEVEL_CHANGE constant being an int, if update.level is returning a long, the constant may need to be changed to long as well.

### DIFF
--- a/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/DatabaseHelper.java
+++ b/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/DatabaseHelper.java
@@ -55,7 +55,7 @@ public final class DatabaseHelper extends Thread {
     private static final double SMALL_LATLON_CHANGE = 0.0001D;
     private static final double MEDIUM_LATLON_CHANGE = 0.001D;
     private static final double BIG_LATLON_CHANGE = 0.01D;
-    private static final int LEVEL_CHANGE = 3;
+    private static final int LEVEL_CHANGE = 5;
     private static final String DATABASE_NAME = "wiglewifi.sqlite";
     private static final String DATABASE_PATH = Environment.getExternalStorageDirectory() + "/wiglewifi/";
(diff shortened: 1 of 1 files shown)


wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/DatabaseHelper.java
```diff
@@ -696,7 +698,7 @@ private void addObservation( final DBUpdate update, final int drainSize ) throws
         //    + " changeWorthy: " + changeWorthy + " mediumChange: " + mediumChange + " smallLocDelay: " + smallLocDelay
         //    + " smallChange: " + smallChange + " latDiff: " + latDiff + " lonDiff: " + lonDiff);
 
-        if ( !blank && (isNew || bigChange || (! fastMode && changeWorthy )) ) {
+        if ( !blank && (isNew || bigChange || (! fastMode && changeWorthy ) || levelChange) ) {
```
bobzilladev: I'd add this to changeWorthy instead, so it's protected by fastMode.

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/DatabaseHelper.java
```diff
@@ -55,6 +55,7 @@
     private static final double SMALL_LATLON_CHANGE = 0.0001D;
     private static final double MEDIUM_LATLON_CHANGE = 0.001D;
     private static final double BIG_LATLON_CHANGE = 0.01D;
+    private static final int LEVEL_CHANGE = 3;
```
bobzilladev: This number feels a bit small, can you switch to 5 and we'll do some testing for impact.

cognitive-rift:
Changes have been made.

PR #98, opened 2017-02-18
ENH - Adds Option to Require GPS Accuracy to Record Data

cognitive-rift:
This adds an option to the "Settings" screen to limit the collection of data points gathered with inaccurate GPS readings.

The Options
High (4m accuracy or Lower)
Medium (10m accuracy or Lower)
Low (20m accuracy or Lower)
Off (All data will be collected) - Default

This is done to both improve quality of data, and reduce export/upload filesize/bandwidth.


cognitive-rift:
This patch is tested as working.

bobzilladev:
I'd rather do this kind filtering server side (which already happens). Even if it is off by a kilometer, if it's the only observation for a network it still shows what country and region the network is in, and collects non-geo data like encryption and channel. Perhaps this can be used as an option to filter the map/list views of the app, instead of filtering what data is recorded?

cognitive-rift:
That's understandable, if the priority is a s many as data points as possible, that's cool. Just looking into the CSV file it seemed wasteful that there would be 5-10 obs for a place sometimes when the server in the end only used the more accurate ones.

PR #102, opened 2017-03-04
ENH - Encrypted token storage

rksh:
First cut at improving token storage using the android keystore system where supported


bobzilladev:
merge away, i'll test on-device.

PR #103, opened 2017-03-07
ENH - styling upload button to match site color

rksh:
avoided square corners so it's not TOO Nat-Geo.


rksh:
Preview Change - adds border stroke in upload, mild background gradient, curved corners

bobzilladev:
21 lines to describe a button. you win this round.

PR #105, opened 2017-03-13
ENH - allow auth right on the settings page.

rksh:
Allows explicit "auth" call from settings page after user password entry.
removing stored password on successful auth.


bobzilladev:
seems legit on a quick scan. my plan is to get this building again locally and take it for a spin on-device, since there's so much change in this release.

PR #106, opened 2017-03-15
BUG - Fixes and cleanup

rksh:
the combination of the last two PRs results in some odd behavior for anonymous uploads. these edits will provide beAnonymous after auth the ability to temporarily anonymize uploads without deauthorizing. This may or may not be a direction we want to go. (the alternative is to clear creds upon setting anonymous, removing all access to profile / stats data, and prevent reversion to authenticated uploads without creds-re-entry, reconfirmation)


bobzilladev:
weeee fun. can't wait to try out all this new goo.

PR #107, opened 2017-03-15
ENH- row alignment cleanup

rksh:
tidying row alignment


bobzilladev:
sure, you can just go forward with UI changes, i can't look at them right now anyawy

PR #110, opened 2017-03-18
FIX - I'd misunderstood marshmallow+ camera permission

rksh:


bobzilladev:
not real fond of this new requesting permission thing android did. +1

Issue #82, opened 2017-01-03
Bad location data in location and network probably due to losing GPS


mulaz:
I have a couple of issues with Wigle on a LG Nexus 4 phone (stock android 5.1.1). I'm not sure if it's device specific or a general issue. 

First and a minor one is that sometimes the "location" table contains wrong location data, eg:

```
sqlite> select _id,level,lat,lon,altitude,accuracy,time from location order by lat desc limit 100;
_id         level       lat         lon         altitude    accuracy        time      
----------  ----------  ----------  ----------  ----------  --------------  ----------
748         -87         Inf         -Inf        0.0         566.0791015625  0         
749         -91         Inf         -Inf        0.0         566.0791015625  0     
750         -86         Inf         -Inf        0.0         566.0791015625  0         
751         -90         Inf         -Inf        0.0         566.0791015625  0         
752         -91         Inf         -Inf        0.0         566.0791015625  0         
4252        -66         Inf         -Inf        0.0         0.924082756042  0         
...
```
"Inf" and "-Inf" are not valid WGS84 coordinates, also time=0 even if it's a valid number isn't OK/real. 
This issue is probably caused by saving data before obtaining a GPS fix or after losing it (not sure). 

There are also lines with valid coordinates (although i think the location is wrong - see below) and time=0.

```
_id         level       lat          lon               altitude    accuracy          time      
----------  ----------  -----------  ----------------  ----------  ----------------  ----------
12103       -85         46.06612801  14.5284741499999  0.0         9.52606964111328  0         
12104       -90         46.06603457  14.52453655       0.0         9.52606964111328  0         
```

This causes issues when trying to draw all the points on a map, but can be avoided by filtering out the invalid location when exporting data from sqlite. 

The bigger issue for me is the "network" table, where last- and best- locations are stored. 

I have visualised a part of best-location data (one line in 'network' table is one dot drawn with "bestlon", "bestlat" coordinates - http://i.imgur.com/DF3hHTc.png ). I have actually walked/scanned the bottom part, but the slightly right leaning horizontal "line of dots" is due to losing GPS/location errors described above. I never actually walked there. 

[...]

PR #111, opened 2017-03-19
FIX - #82 and numerous other error reports.

rksh:
https://github.com/wiglenet/wigle-wifi-wardriving/issues/82

### DIFF
--- a/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/DatabaseHelper.java
+++ b/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/DatabaseHelper.java
@@ -694,12 +694,15 @@ public final class DatabaseHelper extends Thread {
                 && location.getAltitude() == 0 && location.getAccuracy() == 0
                 && update.level == 0;
 
-        /*
-        ALIBI: debugging path
+        /**
+         * ALIBI: +/-infinite lat/long, 0 timestamp data (even with high accuracy) is gigo
+         */
         final boolean likelyJunk = Double.isInfinite(location.getLatitude()) ||
                 Double.isInfinite(location.getLongitude()) ||
                 location.getTime() == 0L;
 
+        /*
+        //debugging path
         if (likelyJunk) {
             MainActivity.info(network.getSsid() + " " + bssid + ") blank: " + blank + "isNew: " + isNew + " bigChange: " + bigChange + " fastMode: " + fastMode
                         + " changeWorthy: " + changeWorthy + " mediumChange: " + mediumChange + " smallLocDelay: " + smallLocDelay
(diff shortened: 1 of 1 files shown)


wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/DatabaseHelper.java
```diff
@@ -694,11 +694,15 @@ private void addObservation( final DBUpdate update, final int drainSize ) throws
...
+        final boolean junk = Double.isInfinite(location.getLatitude()) ||
+                Double.isInfinite(location.getLongitude()) ||
+                location.getTime() == 0L;
+
         // MainActivity.info(network.getSsid() + " " + bssid + ") blank: " + blank + "isNew: " + isNew + " bigChange: " + bigChange + " fastMode: " + fastMode
         //    + " changeWorthy: " + changeWorthy + " mediumChange: " + mediumChange + " smallLocDelay: " + smallLocDelay
         //    + " smallChange: " + smallChange + " latDiff: " + latDiff + " lonDiff: " + lonDiff);
 
-        if ( !blank && (isNew || bigChange || (! fastMode && changeWorthy )) ) {
+        if ( !blank && !junk && (isNew || bigChange || (! fastMode && changeWorthy )) ) {
```
bobzilladev: I still don't want to drop data at the client, ever. Netid, ssid, crypto, etc are all still valid. With this change, if the only observation the client has meets the junk criteria it will never be recorded or sent to the server.

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/DatabaseHelper.java
```diff
@@ -746,7 +750,14 @@ private void addObservation( final DBUpdate update, final int drainSize ) throws
...
 
-                boolean newBest = bestlevel == 0 || update.level > bestlevel;
+                boolean newBest = (bestlevel == 0 || update.level > bestlevel) &&
+                        // https://github.com/wiglenet/wigle-wifi-wardriving/issues/82
+                        // ALIBI: frequent high-accuracy + time = 0
+                        ((location.getTime() != 0L) &&
+                                // ALIBI: frequent high-accuracy + infinite lat/long
+                                !Double.isInfinite(location.getLatitude()) &&
+                                !Double.isInfinite(location.getLongitude()) );
+
```
bobzilladev: Not setting newBest for the junk criteria makes sense. As implemented, it could never get here, but could if you remove the !junk in the if block above, which I would advocate to remove.

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/DatabaseHelper.java
rksh: i agree in principal, and I get how doing existence checks would be expensive...  but does it make sense to dump junk observations if there's a known record for this net already and one of the "junk" tests are met? Experiments with the new KML generation criteria on the server side suggest that many of these records have viable, better records available.

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/DatabaseHelper.java
rksh: figured this would be the preferred approach.

bobzilladev:
When calculating newBest could use likelyJunk instead of computing it again.

PR #112, opened 2017-03-22
FIX - fix for location on map

rksh:
invert perms check so we can get position when allowed ;)

### DIFF
--- a/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/MappingFragment.java
+++ b/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/MappingFragment.java
@@ -155,7 +155,7 @@ public final class MappingFragment extends Fragment {
             public void onMapReady(final GoogleMap googleMap) {
                 if (ActivityCompat.checkSelfPermission(MappingFragment.this.getContext(),
                         android.Manifest.permission.ACCESS_FINE_LOCATION) == PackageManager.PERMISSION_GRANTED
-                        && ActivityCompat.checkSelfPermission(MappingFragment.this.getContext(),
+                        || ActivityCompat.checkSelfPermission(MappingFragment.this.getContext(),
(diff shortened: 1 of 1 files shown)


wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/MappingFragment.java
```diff
@@ -154,9 +154,9 @@ private void setupMapView(final View view, final LatLng oldCenter, final int old
             @Override
             public void onMapReady(final GoogleMap googleMap) {
                 if (ActivityCompat.checkSelfPermission(MappingFragment.this.getContext(),
-                        android.Manifest.permission.ACCESS_FINE_LOCATION) != PackageManager.PERMISSION_GRANTED
+                        android.Manifest.permission.ACCESS_FINE_LOCATION) == PackageManager.PERMISSION_GRANTED
                         && ActivityCompat.checkSelfPermission(MappingFragment.this.getContext(),
```
bobzilladev: && needs to be || as well

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/MappingFragment.java
```diff
@@ -155,7 +155,7 @@ private void setupMapView(final View view, final LatLng oldCenter, final int old
             public void onMapReady(final GoogleMap googleMap) {
                 if (ActivityCompat.checkSelfPermission(MappingFragment.this.getContext(),
                         android.Manifest.permission.ACCESS_FINE_LOCATION) == PackageManager.PERMISSION_GRANTED
-                        && ActivityCompat.checkSelfPermission(MappingFragment.this.getContext(),
+                        || ActivityCompat.checkSelfPermission(MappingFragment.this.getContext(),
```
rksh: per feedback. is it even possible to set these independently at the app level (obvi the system-level permissions are common)

bobzilladev:
Pry not usually right now, we always ask for fine, which gets coarse along for free. But folks can go into settings and change perms anytime now a days.

PR #113, opened 2017-03-24
FIX - marshmallow + toast + permissions dialog = fail

rksh:
seeing complaints from users in appstore re: 2.16 lockout due to "overlay detection" during initial permissions requests.


bobzilladev:
sigh, this is so stupid. +1

PR #114, opened 2017-03-27
FIX - clearing if we can't set

rksh:
saving us from getting into an inconsistent state at the expense of user experience - we'll nuke the token and force re-authentication


bobzilladev:
The idea is sound i think, we don't want to get in a situation where a user can't get out of it. I'd beat the copy+paste drum for this though :)

bobzilladev:
krad++

PR #115, opened 2017-03-28
ENH- don't let the OS optimize battery usage

rksh:
Believe the frequent "wifi resets" and scan lag in android Marshmallow and higher are a symptom of battery optimization - will be testing extensively before merging this fix.

There's some inconclusive documentation that suggests that requesting and using this permission can result in app suspension from the Play Store.

### DIFF
--- a/wiglewifiwardriving/src/main/AndroidManifest.xml
+++ b/wiglewifiwardriving/src/main/AndroidManifest.xml
@@ -35,6 +35,8 @@
     <!-- confine camera activation to Marshmallow and up. -->
     <!-- NB: this functionality exists as early as JellyBean (4.2.2), but not worth the thrash -->
     <uses-permission-sdk-23 android:name="android.permission.CAMERA" />
+    <!-- scan lag due to Marshmallow-and-upBattery Optimization -->
+    <uses-permission-sdk-23 android:name="android.permission.REQUEST_IGNORE_BATTERY_OPTIMIZATIONS" />
 
     <permission
         android:name="net.wigle.wigleandroid.permission.MAPS_RECEIVE"
--- a/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/MainActivity.java
+++ b/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/MainActivity.java
@@ -333,6 +333,18 @@ public final class MainActivity extends AppCompatActivity {
             }
             addPermission(permissionsList, Manifest.permission.WRITE_EXTERNAL_STORAGE);
 
+            PowerManager pm = (PowerManager) this.getSystemService(Context.POWER_SERVICE);
+            if (pm.isIgnoringBatteryOptimizations(this.getPackageName())) {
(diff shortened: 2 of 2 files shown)


rksh:
DO NOT MERGE UNTIL VERIFICATION OF FIX IS COMPLETE, AND UNTIL WE HAVE CONFIRMATION THAT THIS WILL NOT RESULT IN PLAY STORE SUSPENSION

bobzilladev:
Interesting. I've never run into this, but have seen the complaints that may be associated.

rksh:
we have anecdotal reports that disabling this helps, however i've managed to trigger the failure mode on a MotoX 2nd gen while the powerManager says idle == false. May be a dead end.

rksh:
further testing suggests that performance problems on some phones are not related to this. Closing PR.

Raffaele74:
Hi. I can confirm. I have gained some experience as a user and support to some developers. I do not think WiGLE has a battery power problem over the thresholds. But this is the first time EMUI asks me to kill because app has excessive demand to upgrade the position GPS.
First app is "Vivibici" but Android doesn't ask for it, but ask me for WiGLE

screenshot_2017-04-30-13-36-08

Raffaele74:
You cannot have a permission "Ignore". you can only ask to users, because it is manual setting.
But also if user set "Ignore"=ok devices that have it (not Samsung, for exemple) don't "Ignore" but ask in notification (that it is worst becasue users notice it!).

Raffaele74:
If you explane me about setting in WiGLE GPS frequency I can test different setting.

Raffaele74:
More general low battery drain is better. Best an app I can do always on. So:

Few measurament is a shame, but too much also: it drains battery and overloads servers.
So I think right would be to do a measure every time I move to a significant distance, for example 60 meters.
If I'm stopped I do not see the utility of doing measurements then I should like put it off.
If we want to put a scale we consider that at 1 Km / h 60 meters I travel in 216 seconds; At 2 Km / h 108 seconds.
At 8 Km/h I put 27 seconds, but it is up to 8 so I should be able to choose at least 50.
To get under the second I have to exceed 145 Km / h. It seems in my mind that I can consider 1 second the extreme threshold.
For the battery I have to make very sure I can call, so 20% seems to me to be minimal; 50% adequate.

If you could make a simple system that reminds phone areas already beaten to not repeat, this would be a great incentive to keep the App on; app would consume few and it becomes an advantage to speed up the coverage of habitual areas. An App I don't must remember to exit (and so always on) is better. If I can do always on is the best.

An action can be insert timetable. So I can put off in the night and at work. What do you think about? 
Another, slightly more difficult, is to be able to define "completed" cell zones.

**Short: what are best ranges (in my mind)?
stopped     setting from off to 240 by 20-30 (i like very much if app does nothing if I was <2-3 Km/h)
\<8            off to 120 by 10-20
\>8            1 to 50 by 3-5 (but also off #)
\> 50         1 to 24 by 2-3
Battery       off to 50% by 5%-10%**

\# I can need to put off habitual movies. It can help to silence home-work. Or if I decide to do systematic wardriving only by walking, as hobby.
50 km/h" threshold helps distinguish case in/out town

PR #120, opened 2017-04-22
FIX - blanket exception handling around JSON elements

rksh:
It's not the right way to handle this, but a production app shouldn't error out because of a sitedown/etc, and we are getting weird error reports.

also, sorry about the build script updates in here, if those are too big a deal, i'll recut.

### DIFF
--- a/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/background/ObservationImporter.java
+++ b/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/background/ObservationImporter.java
@@ -86,6 +86,10 @@ public class ObservationImporter extends AbstractProgressApiRequest {
                 e.printStackTrace();
                 status = Status.EXCEPTION;
                 bundle.putString(BackgroundGuiHandler.ERROR, "ex problem: " + e);
+            } catch (Exception e) {
+                e.printStackTrace();
+                status = Status.EXCEPTION;
+                bundle.putString(BackgroundGuiHandler.ERROR, "ex problem: " + e);
             } finally {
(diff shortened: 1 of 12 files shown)


wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/background/ObservationImporter.java
```diff
@@ -86,6 +86,10 @@ protected void subRun() throws IOException, InterruptedException {
                 e.printStackTrace();
                 status = Status.EXCEPTION;
                 bundle.putString(BackgroundGuiHandler.ERROR, "ex problem: " + e);
+            } catch (Exception e) {
+                e.printStackTrace();
+                status = Status.EXCEPTION;
+                bundle.putString(BackgroundGuiHandler.ERROR, "ex problem: " + e);
```
uhtu: is it worth differentiating this case? identical handling suggests folding up with the :86 case or altering the putstring?

uhtu:
the more-generic-catch fallback does not seem like a bad handling pattern in general for java's wacky heirarchy

bobzilladev:
so deliciously lame (that we need to do this)! i mean...ENTERPRIZE. +1   :)

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/background/ObservationImporter.java
rksh: yeah. imo, since we might need to debug new horrors here, I'd like to leave a different case, but i'll get more info into that one.

uhtu:
+1

PR #124, opened 2017-04-28
FIX - Hide register link

rksh:
performing activation unsets "be anonymous" (and thereby removes the activate link) bug report from Raffaele.


bobzilladev:
lgtm

PR #125, opened 2017-05-01
Update strings.xml

Raffaele74:


bobzilladev:
This is the default english strings, for a given language you should keep editing the strings with the iso code in the values path.

PR #126, opened 2017-05-01
Add 24 lines and translate

Raffaele74:
Hi Bob. Sorry for last achanges today 4, but "it will view"="mostra" and "it is viewing"="mostra", so I changed "Mostra" because in the app it is not always clear is it is "as is" or "to be".
In filter it is clear because you use box.
It is low priority.


Raffaele74:
Hi Bob. Sorry for change, but "it will view"="mostra" and "it is viewing"="mostra", so I cgange "Mostra" because in the app it is not always clear is it is "as is" or "to be".
In filter it is clear because you use box.
It is low priority.

Raffaele74:
Changed strings too long.

bobzilladev:
Thanks again!

PR #137, opened 2017-05-20
FIX- handling two bug reports from email

rksh:
there are probably a lot more exceptions possible in the uploader case. :/


bobzilladev:
This still looks like it is eating the error, since it's not assigned anywhere. How does this error get shown to the user?

rksh:
Are you advocating for displaying the actual exception to the user? The null returns here both fail gracefully - showing unable-to-connect messages instead of dumping stack and crashing the app.

bobzilladev:
This is better than crashing, but yes, i'd argue showing the error to the user. I'd like to make an activity where the user can see the last 20 lines of logging that i now keep around so they could at least investigate (or we could ask them to send it).

bobzilladev:
Should we close this, since it sounds like these are already not app-crashing? The real fix is probably the flag saying whether this was an app-crashing stack?

PR #138, opened 2017-05-25
ENH - Inline progress

rksh:
This got way, way too big.
- adds hdpi and xxhdpi encryption icon support
- changes myobserved download processing to jackson from obj json (memory footprint fix)
- converts progress dialogs (depr. in android O) to inline progress at bottom of screen
- retrofits UI to prevent more than one large UL/DL at once
- partial updates to language packs


wiglewifiwardriving/build.gradle
```diff
@@ -16,6 +16,15 @@ android {
...
     }
+
+    packagingOptions {
+        exclude 'META-INF/ASL2.0'
+        exclude 'META-INF/LICENSE'
+        exclude 'META-INF/NOTICE'
+        exclude 'META-INF/NOTICE.txt'
+        exclude 'META-INF/LICENSE.txt'
+        exclude 'META-INF/MANIFEST.MF'
+    }
```
rksh: copying these across from the gradle import breaks build - we need our META-INF, not our library

wiglewifiwardriving/build.gradle
```diff
@@ -25,4 +34,7 @@ dependencies {
     compile 'com.google.android.gms:play-services-vision:9.4.0'
     compile 'org.slf4j:slf4j-android:1.7.19'
     compile files('libs/TTS_library_stub.jar')
+    compile 'com.fasterxml.jackson.core:jackson-core:2.9.0.pr3'
+    compile 'com.fasterxml.jackson.core:jackson-annotations:2.9.0.pr3'
+    compile 'com.fasterxml.jackson.core:jackson-databind:2.9.0.pr3'
```
rksh: adding recent jackson version for streaming JSON

wiglewifiwardriving/src/main/res/values/strings.xml
```diff
@@ -333,10 +333,12 @@
...
     <string name="show_discovered_since">Show Networks Discovered Since</string>
+    <string name="show_discovered">Show Discovered Networks on Map</string>
     <string name="map_none">None</string>
-    <string name="map_mine">Mine</string>
-    <string name="map_not_mine">Others</string>
     <string name="map_all">All</string>
+    <string name="map_not_mine">Others</string>
+    <string name="map_mine">Mine</string>
+    <string name="status_downloading">Downloading</string>
+    <string name="status_parsing">Parsing</string>
```
rksh: this is what comes of trusting the UI in android studio to keep these organized.

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/DataFragment.java
```diff
@@ -241,6 +241,10 @@ public void onClick( final View buttonView ) {
 
     private void setupImportObservedButton( final View view ) {
         final Button importObservedButton = (Button) view.findViewById( R.id.import_observed_button );
+        //TODO: is this the right way to use primitives/MA.state?
```
bobzilladev: pry fine, state is used for when we have to rebuild the entire app due to rotating the screen, another awesome androidism.

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/background/ObservationImporter.java
```diff
@@ -36,62 +39,147 @@ public ObservationImporter(final FragmentActivity context,
...
+                    bundle.putString(BackgroundGuiHandler.ERROR, "Connection problem: " + e);
+                } catch (Exception e) {
+                    e.printStackTrace();
+                    status = Status.EXCEPTION;
+                    bundle.putString(BackgroundGuiHandler.ERROR, "ERROR: " + e + " (from " + e.getCause()+")");
+                } finally {
+                    listener.requestComplete(null, false);
+                }
+            } else {
+                //TODO: old code, eliminate (pending confirmation with bobzilla)
```
bobzilladev: This was the entire old code, so if you replaced it then i don't know why you'd still have it. when can cacheFilename be null? it could on the old path.

bobzilladev:
seems ok

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/background/ObservationImporter.java
rksh: the problem here is that without writing to file / reading from file in this specific case, we'll OOM more and more as individual observations-lists get larger. cachefile name is hardcoded, but I'll double check.

bobzilladev: in current mode where it downloads everything from server, i think cachefile is required. i always meant to turn this into a chunked thing, so you you always just got n records, and could then request more. if we're just inserting n at a time it's possible to do that streaming right into the db. in current world, forcing it to always cache to disk is pry the right move.

rksh: to the future!

PR #141, opened 2017-05-27
ENH - adding a cancel button for inline progress

rksh:
adding a simple cxl button in the inline progress dialog. still results in exception dialogs on cancel (may be delayed in some cases). will need visual refinement, testing


bobzilladev:
let's do it!

PR #151, opened 2017-07-09
ENH - new download handling for observation importer.

rksh:
streaming JSON, some cleanups with auth. need to make sure that these are all kosher.

### DIFF
--- a/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/DataFragment.java
+++ b/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/DataFragment.java
@@ -239,18 +239,23 @@ public final class DataFragment extends Fragment implements ApiListener, Transfe
 
     private void setupImportObservedButton( final View view ) {
         final Button importObservedButton = (Button) view.findViewById( R.id.import_observed_button );
-        if (MainActivity.getMainActivity().isTransferring()) {
-            importObservedButton.setEnabled(false);
-        }
+        final SharedPreferences prefs = getActivity().getSharedPreferences(ListFragment.SHARED_PREFS, 0);
+        final String authname = prefs.getString(ListFragment.PREF_AUTHNAME, null);
 
-        importObservedButton.setOnClickListener( new OnClickListener() {
-            @Override
-            public void onClick( final View buttonView ) {
-            MainActivity.createConfirmation(getActivity(),
-                    DataFragment.this.getString(R.string.data_import_observed),
-                    MainActivity.DATA_TAB_POS, IMPORT_DIALOG);
-            }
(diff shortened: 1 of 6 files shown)


wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/DataFragment.java
```diff
@@ -239,18 +239,23 @@ public void onClick( final View buttonView ) {
...
-        if (MainActivity.getMainActivity().isTransferring()) {
+        final SharedPreferences prefs = getActivity().getSharedPreferences(ListFragment.SHARED_PREFS, 0);
+        final String authname = prefs.getString(ListFragment.PREF_AUTHNAME, null);
+
+        if (null == authname) {
             importObservedButton.setEnabled(false);
+        } else if (MainActivity.getMainActivity().isTransferring()) {
+                importObservedButton.setEnabled(false);
+        } else {
+            importObservedButton.setOnClickListener(new OnClickListener() {
```
bobzilladev: Pry would leave the listener outside the else, so it's always set, so if the button is enabled things will work.

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/background/ApiDownloader.java
```diff
@@ -33,11 +34,10 @@ protected void subRun() throws IOException, InterruptedException, WiGLEAuthExcep
             }
             final JSONObject json = new JSONObject(result);
             listener.requestComplete(json, false);
-        }
-        catch (final WiGLEAuthException waex) {
+        } catch (final WiGLEAuthException waex) {
             // ALIBI: allow auth exception through
             throw waex;
-        } catch (final Exception ex) {
+        } catch (final JSONException ex) {
```
bobzilladev: Is there a reason to reduce scope of exception? hopefully handled higher up?

bobzilladev:
yay streaming! lgtm, only minor comments.

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/DataFragment.java
rksh: roger that.

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/background/ApiDownloader.java
rksh: general purpose exception handler was preventing error bubbling up - really just needed JSONException handling at this level

bobzilladev:
+1

PR #153, opened 2017-07-31
FIX - lineageOS token lockout

rksh:
this may not be the best solution, but if we can't access the token, I suppose we should clear it?

### DIFF
--- a/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/TokenAccess.java
+++ b/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/TokenAccess.java
@@ -226,8 +226,16 @@ public class TokenAccess {
                 }
             } catch (CertificateException | NoSuchAlgorithmException | IOException |
                     KeyStoreException | UnrecoverableEntryException | NoSuchPaddingException |
-                    InvalidKeyException | BadPaddingException | IllegalBlockSizeException ex) {
+                    InvalidKeyException | BadPaddingException | IllegalBlockSizeException |
+                    UnsupportedOperationException ex) {
                 MainActivity.error("[TOKEN] Failed to get API Token: ", ex);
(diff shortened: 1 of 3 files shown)


wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/TokenAccess.java
```diff
@@ -226,8 +226,16 @@ public static String getApiToken(SharedPreferences prefs) {
...
+                    InvalidKeyException | BadPaddingException | IllegalBlockSizeException |
+                    UnsupportedOperationException ex) {
                 MainActivity.error("[TOKEN] Failed to get API Token: ", ex);
+                /**
+                 * ALIBI: in the event of failure, we'll clear, force a re-auth
+                 */
+                clearApiToken(prefs);
+                final SharedPreferences.Editor editor = prefs.edit();
+                editor.remove(ListFragment.PREF_AUTHNAME);
+                editor.apply();
```
bobzilladev: i think this is good behavior, as long as it lets us clear in this situation. no idea how to test that. one thing though, can we let the user know this happened? just a toast or something?

rksh: how about handling this via ObservationUploader below?

bobzilladev:
wfm +1

rksh:
please do not merge yet - trying to find a way to confirm that this is sufficient to fix.

bobzilladev:
Will the aes change fix things? Or would you like to de-conflict this?

rksh:
doesn't seem particularly useful compared to https://github.com/wiglenet/wigle-wifi-wardriving/pull/157 - closing, we can re-add later if we're having trouble after AES conversion.

PR #154, opened 2017-08-01
ENH - allowing a broadcast signal-based app termination

rksh:
from https://wigle.net/phpbb/viewtopic.php?f=3&p=9317#p9317

also removes some old intent handling code in MainActivity
tested using Tasker and adb.

### DIFF
--- a/wiglewifiwardriving/src/main/AndroidManifest.xml
+++ b/wiglewifiwardriving/src/main/AndroidManifest.xml
@@ -72,18 +72,6 @@
                 <action android:name="android.intent.action.MAIN" />
                 <category android:name="android.intent.category.LAUNCHER" />
             </intent-filter>
-            <intent-filter>
-                <action android:name="android.intent.action.ACTION_SYNC" />
-                <category android:name="android.intent.category.DEFAULT" />
-            </intent-filter>
-            <intent-filter>
-                <action android:name="android.intent.action.ACTION_INSERT" />
-                <category android:name="android.intent.category.DEFAULT" />
-            </intent-filter>
-            <intent-filter>
-                <action android:name="android.intent.action.ACTION_DELETE" />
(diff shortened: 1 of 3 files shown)


wiglewifiwardriving/src/main/AndroidManifest.xml
```diff
@@ -72,18 +72,6 @@
...
-                <category android:name="android.intent.category.DEFAULT" />
-            </intent-filter>
-            <intent-filter>
-                <action android:name="android.intent.action.ACTION_INSERT" />
-                <category android:name="android.intent.category.DEFAULT" />
-            </intent-filter>
-            <intent-filter>
-                <action android:name="android.intent.action.ACTION_DELETE" />
-                <category android:name="android.intent.category.DEFAULT" />
-            </intent-filter>
```
bobzilladev: I believe these are needed for the service-notifcation pause/scan buttons, no?

rksh: I thought these were the intent targets that got moved to ShareActivity per https://github.com/wiglenet/wigle-wifi-wardriving/blob/master/wiglewifiwardriving/src/main/AndroidManifest.xml#L175. WiGLEService maps them through ShareActivity at https://github.com/wiglenet/wigle-wifi-wardriving/blob/master/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/WigleService.java#L199  - seems to work in testing

rksh: (tested both from ListFragment menus and status badge pause/scan buttons)

bobzilladev: Ah, ok, looks right. as long as that's tested to still work. +1!

PR #156, opened 2017-08-05
ENH - filter list dialog -> activity

rksh:
mostly plumbing to fit the filter screen on all devices (preparation to grow the set of list filters per feature req). Does not address trouble with map filter dialog in some languages.


bobzilladev:
super +1

PR #158, opened 2017-08-07
FIX - creating a gingerbread-friendly Legacy observation import.

rksh:
inefficient, but will work for small sets - returns big JSON string, then parses it all in memory (like the bad old days).
This could should be removed as soon as we drop Gingerbread support.


bobzilladev:
looks legit. assume this is pasting in the old bits from before. as long as it's tested, +1

rksh:
yep, just bringing the old code back, with a small clean-up in imports.

PR #159, opened 2017-08-09
ENH - map filter and tiles as an activity from Map Fragment

rksh:
first cut at changing map settings to own dialog, combining with maptile settings.

template for clarifying main settings mage, modularizing settings that appear in more than one place.

### DIFF
--- a/wiglewifiwardriving/src/main/java/android/support/v4/app/FixedMapSavedState.java
+++ b/wiglewifiwardriving/src/main/java/android/support/v4/app/FixedMapSavedState.java
@@ -1,0 +1,26 @@
+package android.support.v4.app;
+
+import android.os.Bundle;
+import android.os.Parcel;
+import android.support.v4.app.Fragment;
+
+/**
+ * Created by arkasha on 7/3/17.
+ */
+
+public class FixedMapSavedState extends Fragment.SavedState {
+    FixedMapSavedState(Bundle b) {
+        super(b);
+    }
+
+
+    FixedMapSavedState(Parcel in, ClassLoader loader) {
+        super(in, loader);
+    }
+
+    public FixedMapSavedState(Fragment.SavedState ss, ClassLoader cl) {
+        super(ss.mState);
+        ss.mState.setClassLoader(cl);
+    }
+}
+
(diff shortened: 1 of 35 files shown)


wiglewifiwardriving/src/main/java/android/support/v4/app/FixedMapSavedState.java
```diff
@@ -0,0 +1,26 @@
...
+
+import android.os.Bundle;
+import android.os.Parcel;
+import android.support.v4.app.Fragment;
+
+/**
+ * Created by arkasha on 7/3/17.
+ */
+
+public class FixedMapSavedState extends Fragment.SavedState {
```
bobzilladev: oh my, what's the story behind this thing?

rksh: having just completed the gingerbread back-compat change, I tried this fix on my Gingerbread Nexus S instance - to find that I hadn't completely fixed a badParcelable exception with maps; this change was a provisional diff I had hanging around from efforts earlier in the month, fixes in emulation.

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/MappingFragment.java
```diff
@@ -87,11 +87,14 @@
     private AtomicBoolean finishing;
     private Location previousLocation;
     private int previousRunNets;
+    private TileOverlay tileOverlay;
```
rksh: shoving this out to the class level so we can mess with it onResume

wiglewifiwardriving/src/main/res/layout/settings.xml
```diff
@@ -256,20 +213,21 @@
...
 		/>
 	<TextView
 		android:id="@+id/gps_text"
-		android:layout_width="wrap_content"
-		android:layout_height="wrap_content"
+		android:layout_width="0dp"
+		android:layout_height="match_parent"
 		android:text="@string/gps_text"
-		android:layout_weight="1"
+		android:layout_weight="2"
```
rksh: so, in languages where "portuguese (brazilian) becomes too wide for the screen, we force the label text into a 2DP column at the far right of the screen. This prevents that.

PR #160, opened 2017-08-10
ENH - organizing the settings screen into sections

rksh:
breaking settings up into four sections to help people parse the giant list
screen shot 2017-08-10 at 3 20 20 pm
screen shot 2017-08-10 at 3 20 37 pm


bobzilladev:
i like the sections! i think map settings should just live on the map fragment itself still, but i don't feel super strongly about it.

rksh:
with the exceptions of the settings that only lived in the main settings menu before (circlesize?) they're all available there too!

PR #162, opened 2017-08-13
Updated danish language

Zarhn:
*Translated untranslated lines.
*Updated bad danish gramman
*Added *.db to gitignore cause windows put *.db files everywere.


bobzilladev:
Thanks!

PR #165, opened 2017-08-22
fix npe ActivateActivity.java L69

bobzilladev:


wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/ShareActivity.java
```diff
@@ -21,18 +21,21 @@ protected void onCreate(Bundle savedInstanceState) {
...
-                break;
-            case Intent.ACTION_DELETE:
-                MainActivity.getMainActivity().handleScanChange(false);
-                break;
-            case Intent.ACTION_SYNC:
-                MainActivity.getMainActivity().doUpload();
-                break;
-            default:
-                MainActivity.info("Unhandled intent action: " + intent.getAction());
+        final MainActivity mainActivity = MainActivity.getMainActivity();
```
rksh: what kind of idiot would skip the nullcheck here?! oh, it's me.

bobzilladev: hehe, I would bet that one was me.

PR #168, opened 2017-09-06
ENH - first version of filtering.

rksh:
So much left to do, but this change is getting hefty.


wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/listener/WifiReceiver.java
```diff
@@ -220,7 +220,11 @@ public void onReceive( final Context context, final Intent intent ) {
 
         final boolean ssidSpeak = prefs.getBoolean( ListFragment.PREF_SPEAK_SSID, false )
                 && ! mainActivity.isMuted();
-        final Matcher ssidMatcher = FilterMatcher.getFilterMatcher( prefs, ListFragment.FILTER_PREF_PREFIX );
+
+        //TODO: should we memoize the ssidMatcher in the MainActivity state as well?
```
rksh: didn't port SSID matcher to centralized, memoized version 1. to minimize code churn and 2. because I wanted to bounce the new impl off of folks before expanding

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/MapRender.java
```diff
@@ -244,7 +244,9 @@ public boolean okForMapTab( final Network network ) {
                 && ! isDbResult;
         if (network.getPosition() != null) {
             if (!showNewDBOnly || network.isNew()) {
-                if (FilterMatcher.isOk(ssidMatcher, prefs, MappingFragment.MAP_DIALOG_PREFIX, network)) {
+                if (FilterMatcher.isOk(ssidMatcher,
+                        null /*ALIBI: we *can* use the filter from the list filter view here ...*/,
+                        prefs, MappingFragment.MAP_DIALOG_PREFIX, network)) {
```
rksh: don't think we want the list view filter to cross-operate in maps, but we could do that.

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/MacFilterActivity.java
```diff
@@ -0,0 +1,112 @@
...
+            Gson gson = new Gson();
+            String serialized = gson.toJson(entries.toArray());
+            MainActivity.info(serialized);
+            final SharedPreferences.Editor editor = prefs.edit();
+            editor.putString(this.filterKey,serialized);
+            editor.apply();
+            input.setText("");
+            MainActivity m = MainActivity.getMainActivity();
+            if (null != m) {
+                //TODO: should we also update on Suspend/Dispose?
```
rksh: kinda my biggest question about this approach - updating on suspend would be belt-and-suspenders

PR #172, opened 2017-09-10
ENH - first pass at on-detail hide/block of network (filter addition)

rksh:
allows users to exclude nets from list or logging via the network detail screen.
includes some visual tightening to fit controls and improve readability


bobzilladev:
looks legit +1

PR #176, opened 2017-09-18
FIX - don't offer button with weird/corrupt bssids

rksh:
careless, but I hadn't imagined bssids being < 7 characters.


rksh:
open to more defensive practices - was surprised to see these < 7 but not null.

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/NetworkActivity.java
```diff
@@ -334,15 +336,18 @@ public void onClick( final View buttonView ) {
             });
         }
 
-        if (hideAddresses.contains(network.getBssid().toUpperCase())) {
+        if ( (null == network.getBssid()) || (network.getBssid().length() < 17) ||
```
bobzilladev: These look to still not be part of the wifi check block.

PR #178, opened 2017-09-21
FIX - updating file access to supported mode in >= Marshmallow 

rksh:
MODE_WORLD_READABLE is depr. in M, fails in N.

### DIFF
--- a/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/MainActivity.java
+++ b/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/MainActivity.java
@@ -894,8 +894,11 @@ public final class MainActivity extends AppCompatActivity {
             return new FileOutputStream(file);
         }
 
-        //noinspection deprecation
-        return context.openFileOutput(filename, Context.MODE_WORLD_READABLE);
+        if (Build.VERSION.SDK_INT >= 23) {
+            return context.openFileOutput(filename, Context.MODE_PRIVATE);
+        } else {
+            return context.openFileOutput(filename, Context.MODE_WORLD_READABLE);
+        }
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/MainActivity.java
```diff
@@ -1561,7 +1568,7 @@ public void playRunNetSound() {
 
     private void setupWifi() {
         // warn about turning off network notification
-        @SuppressWarnings("deprecation")
+        //@SuppressWarnings("deprecation")
```
bobzilladev: can pry delete

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/MainActivity.java
```diff
@@ -894,8 +894,11 @@ public static FileOutputStream createFile(final Context context, final String fi
             return new FileOutputStream(file);
         }
 
-        //noinspection deprecation
-        return context.openFileOutput(filename, Context.MODE_WORLD_READABLE);
+        if (Build.VERSION.SDK_INT >= 23) {
+            return context.openFileOutput(filename, Context.MODE_PRIVATE);
+        } else {
+            return context.openFileOutput(filename, Context.MODE_WORLD_READABLE);
+        }
```
bobzilladev: I'd do a ternary to chose the mode to a variable, then pass that variable to a single openFileOutput, to reduce copy paste. i don't know if android developer inspection handles that properly though.

rksh: yeah, i was satisfying the noodly editor validation logic. i'll read up on the right way to do it.

bobzilladev:
as long as this doesn't break upgrades over existing files, and can still get to sd files, i'm +100

PR #179, opened 2017-09-21
FIX - toast replacement, suppress in Android 7.1

rksh:
FOR DISCUSSION: heavy-handed, but without a way to reproduce the 7.1 BadTokenExceptions simply, this may be a way...

### DIFF
--- a/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/util/WiGLEToast.java
+++ b/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/util/WiGLEToast.java
@@ -1,0 +1,71 @@
+package net.wigle.wigleandroid.util;
+
+import android.app.Activity;
+import android.os.Build;
+import android.support.v4.app.FragmentActivity;
+import android.view.Gravity;
+import android.view.LayoutInflater;
+import android.view.View;
+import android.view.ViewGroup;
+import android.widget.TextView;
+import android.widget.Toast;
+
+import net.wigle.wigleandroid.MainActivity;
+import net.wigle.wigleandroid.R;
+
+/**
+ * Display that magic WiGLE toast
+ * Created by arkasha on 9/21/17.
+ */
+
+public class WiGLEToast {
+
+    public static void showOverFragment(final FragmentActivity context, final int titleId,
+                                        final String messageString) {
+        LayoutInflater inflater = context.getLayoutInflater();
+        View layout = inflater.inflate(R.layout.wigle_detail_toast,
+                (ViewGroup) context.findViewById(R.id.custom_toast_container));
+
+        TextView title = (TextView) layout.findViewById(R.id.toast_title_text);
+        title.setText(titleId);
+
+        TextView text = (TextView) layout.findViewById(R.id.toast_message_text);
+        text.setText(messageString);
+
+        Toast toast = new Toast(context.getApplicationContext());
+        toast.setGravity(Gravity.CENTER_VERTICAL, 0, 0);
+        toast.setDuration(Toast.LENGTH_LONG);
+        toast.setView(layout);
+        toast.show();
+    }
+
+    public static void showOverActivity(final Activity context, final int titleId,
+                                        final String messageString) {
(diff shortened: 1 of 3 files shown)


wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/util/WiGLEToast.java
```diff
@@ -0,0 +1,71 @@
...
+
+    public static void showOverActivity(final Activity context, final int titleId,
+                                        final String messageString) {
+        showOverActivity(context, titleId,
+                messageString, Toast.LENGTH_SHORT);
+    }
+
+    public static void showOverActivity(final Activity context, final int titleId,
+        final String messageString, final int toastLength) {
+        if (Build.VERSION.SDK_INT != 25) {
```
bobzilladev: should pry be < to be future proof

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/util/WiGLEToast.java
```diff
@@ -0,0 +1,71 @@
...
+            TextView text = (TextView) layout.findViewById(R.id.toast_message_text);
+            text.setText(messageString);
+
+            Toast toast = new Toast(context.getApplicationContext());
+            toast.setGravity(Gravity.CENTER_VERTICAL, 0, 0);
+            toast.setDuration(toastLength);
+            toast.setView(layout);
+            toast.show();
+        } else {
+            MainActivity.info("toast disabled because 7.1.x bombs: "+messageString);
```
bobzilladev: is a plan to make this shown to the users somehow someday again?

bobzilladev: can this use the same area overlay as the cool upload progress bar?

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/util/WiGLEToast.java
rksh: we have no evidence that it happens in 7.1.2 or 8.0, ne?

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/util/WiGLEToast.java
rksh: this is using that same fancy-toast! (and yes, i'd prefer to show it in as many cases as possible)

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/util/WiGLEToast.java
bobzilladev: er, oh, weird. play console does confirm only happens in that one exact version. wtf. i stand corrected.

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/util/WiGLEToast.java
bobzilladev: it won't be truly fancy toast until Hello Kitty is burned onto it.

rksh: I look forward to the hello kitty PR and will merge it gladly

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/util/WiGLEToast.java
rksh: this is why this is a controversial solution and a privisional PR - I quite like the toast, and anecdotal evidence suggests that it's good scaffolding. Oreo and 7.1.2 installed bases are still small. I presume that this is a patch to prevent the malicious toast crisis (and that that's 7.1 specific), but I can't prove that because I can't reproduce it.

bobzilladev:
I'm totally fine with this. One sdk version in the past can't get toast, so be it.

PR #180, opened 2017-09-29
ENH - Clear db button

rksh:
per user requests, allow clearing of database


bobzilladev:
yikes. wonder if we should make that even harder to do accidently.

rksh:
a fair question; maybe there's a way to surround the dialog with yellow-and-black stripes? I could also paint the button red as a warning... https://i.imgur.com/WYQee.gif

bobzilladev:
heh, exactly. shiny, red, candy-like button.

PR #182, opened 2017-10-07
FIX - Miscategorized errors

rksh:
Struggling to find the source of the problems people are seeing with transient login failure. In this instance we were incorrectly describing connection problems as login problems, which this PR fixes.


wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/background/BackgroundGuiHandler.java
```diff
@@ -62,6 +63,13 @@ public void handleMessage( final Message msg ) {
                     pp.hide();
                 }
             }
+            if (msg.what == CONNECTION_ERROR) {
+                Toast.makeText(this.context, R.string.no_wigle_conn
+                        , Toast.LENGTH_LONG).show();
```
bobzilladev: oh no! moar toast!  :)

PR #183, opened 2017-10-10
FIX - avoiding problems with manually def'd strings.

rksh:
building out a safe index for the connection error problem

### DIFF
--- a/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/background/AbstractBackgroundTask.java
+++ b/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/background/AbstractBackgroundTask.java
@@ -24,6 +24,8 @@ import java.io.IOException;
 import java.util.Locale;
 import java.util.concurrent.atomic.AtomicBoolean;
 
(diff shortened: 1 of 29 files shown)


wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/background/AbstractBackgroundTask.java
```diff
@@ -94,7 +96,7 @@ public final void run() {
             Bundle errorBundle = new Bundle();
             errorBundle.putString(BackgroundGuiHandler.ERROR, "IOException");
             errorBundle.putCharSequence("CONN_ERROR", ioex.getMessage());
-            sendBundledMessage(BackgroundGuiHandler.CONNECTION_ERROR, errorBundle);
+            sendBundledMessage(CONNECTION_ERROR.ordinal(), errorBundle);
```
bobzilladev: maybe should change the other one to that enum too? is weird to send with two different constant types.

rksh:
obviated by simpler PR.

PR #184, opened 2017-10-10
FIX - alternate fix for error messages on upload

rksh:
doesn't get much simpler than this.


rksh:
inclined to leave it alone, since it's been working so far!

PR #185, opened 2017-10-10
FIX - bug report from user about interruption

rksh:
There may still be some cases where a transfer has completed, but tear down gets called anyway.


rksh:
this won't kill the HttpUrlConnections (they're like cockroaches) - but there's no reason not to fix this on its own, I guess.

PR #186, opened 2017-10-14
ENH- making the scanning status indicator into a button.

rksh:
from user requests

### DIFF
--- a/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/ListFragment.java
+++ b/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/ListFragment.java
@@ -27,7 +27,7 @@ import android.widget.AdapterView;
 import android.widget.AdapterView.OnItemSelectedListener;
 import android.widget.ArrayAdapter;
 import android.widget.Button;
-import android.widget.ImageView;
+import android.widget.ImageButton;
 import android.widget.ListView;
 import android.widget.Spinner;
 import android.widget.TextView;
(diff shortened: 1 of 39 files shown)


bobzilladev:
only thought is hitting that by accident would be sad for people hours later, might want a confirmation dialog?

rksh:
interesting point - trade-off between "quick button" and "oops" seems complicated. could warn the first time. or make it preference-able...

bobzilladev:
maybe quick press has a confirm dialog, which also explains that a long-press does it without a dialog?

rksh:
the votes are in, and it's a 3-way tie! :(
https://wigle.net/phpbb/viewtopic.php?f=1&t=2484
waiting for more input.

bobzilladev:
I like the "dialog with checkbox to not alert again" pattern you suggested. Seems to make everyone happy.

rksh:
implemented as discussed

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/ListFragment.java
```diff
@@ -150,6 +152,10 @@
     public static final String ANONYMOUS = "anonymous";
     public static final String WIFI_LOCK_NAME = "wigleWifiLock";
 
+    public static final String QUCIK_SCAN_UNSET = "UNSET";
+    public static final String QUCIK_SCAN_DO_NOTHING = "DO_NOTHING";
+    public static final String QUCIK_SCAN_PAUSE = "PAUSE";
```
bobzilladev: is QUCIK a reference i don't get?

rksh:
closing because rebase crap is in the PR (and shouldn't be)

PR #187, opened 2017-10-17
FIX - DESTROY ALL REMAINING TOAST

rksh:
because android 7.1 just will not stop failing with BadTokenExceptions in odd places.
New strings entries are because of 1. new toast dialogs requiring titles 2. the fact that the odd camera lib-but-no-supported-cam error had not been i18n'd!


bobzilladev:
i__m_making___toast_by_lilostitchfan

PR #192, opened 2017-10-22
FIX - why not just prefer AES?

rksh:
I'm confused about how this worked at all.

### DIFF
--- a/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/TokenAccess.java
+++ b/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/TokenAccess.java
@@ -135,24 +135,29 @@ public class TokenAccess {
             UnrecoverableEntryException, IllegalBlockSizeException, BadPaddingException,
             InvalidAlgorithmParameterException {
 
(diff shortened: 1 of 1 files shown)


bobzilladev:
i think this is ok. all this code is thorny

rksh:
this is our problem with LineageOS - we're seeing differing cipher compat in non-matching versions to the android mainline. getApiTokenVersion2 will never be called *unless* there's a keystore KEYSTORE_WIGLE_CREDS_KEY_V2 set - this isn't a proxy for an M check, but if it ...CREDS_KEY_V2 is set, then it was set in a version where we successfully generated a v2 keypair, which is a "good enough" test. same goes for KEYSTORE_WIGLE_CREDS_KEY_V1 and KEYSTORE_WIGLE_CREDS_KEY_V0 in order - if none are present, then we're safely in a plaintext situation. We never unset old aliases.

bobzilladev:
Then even more-so it needs to be lined up, so either both places do the M check, or neither. With this PR only getApiTokenVersion2 is doing the M check, and if it fails it returns null, even if the key was actually set in the keystore.

rksh:
Relaxed check to K - below that it will generate compiler warnings/errors (GCMParameterSpec not in < v19)

PR #193, opened 2017-10-23
updated German translation

kill0rz:
updated once again :)


kill0rz:
I guess in 343 you did not mean "mine" in sense of to tunnel, which translate into German "Bergwerk" :D

PR #195, opened 2017-10-24
ENH - simplify map control / auto-follow

rksh:
update map lock-on status when users move the map or tap the "my location" button. lots of experimentation to find a minimal solution, but turned out to be really simple.


wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/MappingFragment.java
```diff
@@ -90,6 +88,8 @@
     private int previousRunNets;
     private TileOverlay tileOverlay;
 
+    private Menu menu;
```
bobzilladev: clever solution! i'd recommend null checks around any access of this, unless you're super sure it'll always be set before the rest.

PR #198, opened 2017-11-19
FIX - bug report on scanning change

rksh:
from user bug:

> When using the indicator/button to turn scanning on and off, the title changes to "Scan Off" and "Scan On". Using the menu doesn't do that. Also, the text indication is backwards. I've attached a couple of screenshots to demonstrate.


bobzilladev:
with your magic animation, this pry isn't needed at all, +1

PR #199, opened 2017-12-22
ENH - maybe a working background upload intent.

rksh:
need to test interactions with enter/exit app, make sure we haven't introduced conflicts in state/uploader.


bobzilladev:
Nifty, if that doesn't explode in exceptions. My attempt at that was failsauce.

rksh:
enhanced to preclude self-interruption, tested.

PR #202, opened 2018-01-08
ENH - using a webview to integrate registration into app

rksh:
first cut at a tighter registration into app - relies on javascript asserted into page and user-agent header.


bobzilladev:
nom, can't wait to play with.

bobzilladev:
lovely

PR #208, opened 2018-03-15
ENH - application icon updates based on new ess vee gees

rksh:
so, we created a big vector pile to replace the old screenshot from the never-released 3d client that has served as the icon for all these year. these should render up more cleanly, and be a little more stylish. Also introducing the new XXXhdpi - finally vin diesel can use the app.


bobzilladev:
so clean! so flat! much design!

PR #211, opened 2018-03-24
ENH - adding a run clock to the dash.

rksh:
this is a feature i've always wanted. open to criticism/reasons not to include.

### DIFF
--- a/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/DashboardFragment.java
+++ b/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/DashboardFragment.java
@@ -1,14 +1,12 @@
 package net.wigle.wigleandroid;
 
 import android.content.Context;
-import android.content.Intent;
 import android.content.SharedPreferences;
 import android.content.res.Configuration;
 import android.location.Location;
 import android.media.AudioManager;
 import android.os.Bundle;
 import android.os.Handler;
-import android.provider.Settings;
 import android.support.v4.app.Fragment;
 import android.view.LayoutInflater;
 import android.view.Menu;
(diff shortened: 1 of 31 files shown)


rksh:
we *can* eliminate all the redundant entries in our translation files, but then we lose line-level comparison!

goldfndr:
Does the optimizer know enough to not make two calls to get the system time?

rksh:
that's a fair question. Activating the second call _should_ be a _very_ unreachable corner-case - Since you can't launch the Activity without successfully setting the start time, this would imply that something had modified (deleted) the shared preference after the app had started. With the preferences get, the "default" parameter will only be requested if the preference isn't available, so it will likely never be called.

goldfndr:
I think you might be wrong. [`getLong`](https://developer.android.com/reference/android/content/SharedPreferences.html#getLong\(java.lang.String,%20long\)) has a `long` as its second parameter. If it was a reference, I'd agree, but since it's a value, I disagree.

https://blog.rapid7.com/2017/01/13/java-8-lazy-argument-evaluation/ says that Java 8 uses eager loading, not lazy loading. ("Java evaluates a method’s arguments eagerly")

rksh:
so, we're actively *not* using the functional interface mentioned in the article, but even before that, sending a method call as a fallback value should not result in execution when the value is present for this call. since the param isn't final in the android source, it would be weird to pre-compute it, but I'll benchmark on live devices to make sure. The intentions of this expression is to provide a fallback method, not to eager-execute the call.

rksh:
performance of the calls either way seems to be dominated by external factors - nothing conclusive even with nanosecond measurements. whatever the case, it's fast! I'll do a new pr memoizing the value in case.

rksh:
https://github.com/wiglenet/wigle-wifi-wardriving/pull/212

PR #212, opened 2018-03-26
FIX - memoizing timestamp after suggestion from @goldfndr

rksh:


bobzilladev:
if you can measure the performance impact of this change i'll give you a cookie. +1

rksh:
i tried for a while even at nanosec scale, it's not significant. Since this doesn't lend itself to clean testing, in-situ tests were always dominated by the external workload in emulation. It's not meaningful, it's just defensive- telling the compiler to not be dumb.

PR #214, opened 2018-03-31
ENH - simplified Chinese translation from Geekeyes

rksh:
- incorporating translation file from Geekeyes w/ attribution per request
- branching language descriptions
- loading three optional Chinese variants into settings menu.

Q: unclear whether we can delete `values-zh/strings.xml` now that these are present; does all auto-detection default correctly in all cases? 

THANKS TO GEEKEYES FOR THE TRANSLATION!


bobzilladev:
bonkers, localization ftw

PR #218, opened 2018-04-19
updated french translation

marmat8951:
Correction of the french translation and added translated string for untranslated ones.

ps: You can ask me whenever you want if you are searching for a french translation.


bobzilladev:
Thanks!

PR #220, opened 2018-04-22
default to showing map tiles. fix restart.

bobzilladev:


wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/MainActivity.java
```diff
@@ -1994,16 +1989,29 @@ public boolean onOptionsItemSelected(final MenuItem item) {
...
-    public void finishSoon(final long finishTimeMillis) {
+    public void finishSoon(final long finishTimeMillis, final boolean restart) {
         MainActivity.info("Will finish in " + finishTimeMillis + "ms");
         new Handler().postDelayed(new Runnable() {
             @Override
             public void run() {
+                final Intent i = getBaseContext().getPackageManager()
+                        .getLaunchIntentForPackage(getBaseContext().getPackageName());
+                i.addFlags(Intent.FLAG_ACTIVITY_CLEAR_TOP);
+
```
rksh: nit: can't the intent before instantiated within the if (restart) test to avoid unnecessary  creation in restart == false case?

bobzilladev: i started like that, but thought it might fail to lookup something that was already finished.

bobzilladev:
hrm, point on leaking location info to the server, maybe that should still be opt-in :(

bobzilladev:
maybe i can just make it easier to find, add an option to the regular dropdown 3-dot menu?

PR #221, opened 2018-04-24
feat: Possibility to start Wigle when the phone boot

marmat8951:
Hey!

I've realized this little functionality.
If you think it would be better to do it another way, please, tell me. And if you think it wold be better if it starts in the background and you know how to do it, please tell me.

### DIFF
--- a/TODO
+++ b/TODO
@@ -51,7 +51,6 @@ set minimum on distance decimal to 2 on dashboard
 on networkactivity record more observations on that network (site survey)
 rssi on kml export
 upload and/or pause on when connected to this network
-start on boot option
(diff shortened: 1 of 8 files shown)


bobzilladev:
fun! thanks!

Firecul:
In values/strings.xml there is a small spelling mistake in the new line added, "lunch Wigle at boot"
Instead of "launch Wigle at boot"
I'm on mobile otherwise I'd have added a commit myself

marmat8951:
see: https://github.com/wiglenet/wigle-wifi-wardriving/pull/222

PR #223, opened 2018-04-27
FIX - Notification now deseapear when swipe-killing the application

marmat8951:
Before this commit, when you swipe out the application from the list, the notification of the service saying how many Wifi network was staying. Now, it disappear, as the service is stopped.


bobzilladev:
Turns out to be more complicated, need to invoke finish on main to get the whole thing to shut down properly. Doing in the service which gets a callback onTaskRemoved when this specific action happens, since doing in onDestroy breaks new app install restart.
Done in: https://github.com/wiglenet/wigle-wifi-wardriving/pull/225

marmat8951:
Seen this. Nice one. Couldn't find that way even if I saw mine was pretty bad...

Great job :1st_place_medal:

Issue #81, opened 2016-11-21
Stopping tts can take very long


kuraz758:
It have tts setup to only say the ssids. But when there are many of them it takes a while to say them all. When I use the ... Menu to silence the app it wants to finish saying all ssids it hasn't told me about yet. When I exit the app it stops immediately but in my case I would have to first silence it then exit then restart the app.

rksh:
fix will be in  v2.61

bobzilladev:
Was fixed back then: https://github.com/wiglenet/wigle-wifi-wardriving/commit/619dc98a2c2815d2620e362d921e4660fc280217

PR #225, opened 2018-05-12
handle when user force kills app

bobzilladev:
Turns out to be more complicated, need to invoke finish on main to get the whole thing to shut down properly. Doing in the service which gets a callback onTaskRemoved when this specific action happens, since doing in onDestroy breaks new app install restart.

### DIFF
--- a/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/MainActivity.java
+++ b/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/MainActivity.java
@@ -990,6 +990,8 @@ public final class MainActivity extends AppCompatActivity {
         } catch (final IllegalArgumentException ex) {
             info("wifiReceiver not registered: " + ex);
         }
+
+        if (state.tts != null) state.tts.shutdown();
     }
 
     @Override
@@ -2109,6 +2111,7 @@ public final class MainActivity extends AppCompatActivity {
         if (state.soundNewPop != null) {
             state.soundNewPop.release();
         }
+        info("MAIN: finish complete.");
 
         super.finish();
     }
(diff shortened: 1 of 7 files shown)


wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/MainActivity.java
```diff
@@ -990,6 +990,8 @@ public void onDestroy() {
         } catch (final IllegalArgumentException ex) {
             info("wifiReceiver not registered: " + ex);
         }
+
+        if (state.tts != null) state.tts.shutdown();
```
rksh: oh, interesting - is this related to https://github.com/wiglenet/wigle-wifi-wardriving/issues/81 ?

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/MainActivity.java
```diff
@@ -2109,6 +2111,7 @@ public void finish() {
         if (state.soundNewPop != null) {
             state.soundNewPop.release();
         }
+        info("MAIN: finish complete.");
```
rksh: what, "finished finishing" too trite for you? ;)

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/MainActivity.java
bobzilladev: oh, no, not sure on 81, we tell the tts to read something it just keeps going, would have to play around with interrupts there. this is just to stop a stacktrace about leaking the service when the app is force killed (which calls onDestroy only, not finish, with this change the service notices the app is gone and calls finish later).

PR #230, opened 2018-06-19
ENH - A first version of enhanced cell scanning.

rksh:
This is for discussion, not direct incorporation - this is a really significant rewrite, and requires a great deal of testing and review.

Falls back on old behavior for SDK 14-16 as well as CDMA Phones.
Uses getAllCellInfo on GSM phones SDK 17 and up, maps carrier names from a new hash.
Can detect multiple cells, but currently discards the majority of information from the getAllCells call since it's largely garbage data.

### DIFF
--- a/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/listener/WifiReceiver.java
+++ b/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/listener/WifiReceiver.java
@@ -61,6 +61,7 @@ import android.telephony.CellSignalStrengthLte;
 import android.telephony.CellSignalStrengthWcdma;
 import android.telephony.NeighboringCellInfo;
 import android.telephony.TelephonyManager;
+import android.telephony.cdma.CdmaCellLocation;
 import android.telephony.gsm.GsmCellLocation;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
@@ -692,9 +693,9 @@ public class WifiReceiver extends BroadcastReceiver {
             // ignore
         } else if ( cellLocation.getClass().getSimpleName().equals("CdmaCellLocation") ) {
             try {
-                final int systemId = (Integer) cellLocation.getClass().getMethod("getSystemId").invoke(cellLocation);
-                final int networkId = (Integer) cellLocation.getClass().getMethod("getNetworkId").invoke(cellLocation);
-                final int baseStationId = (Integer) cellLocation.getClass().getMethod("getBaseStationId").invoke(cellLocation);
+                final int systemId = ((CdmaCellLocation) cellLocation).getSystemId();
+                final int networkId = ((CdmaCellLocation) cellLocation).getNetworkId();
+                final int baseStationId = ((CdmaCellLocation) cellLocation).getBaseStationId();
                 if ( systemId > 0 && networkId >= 0 && baseStationId >= 0 ) {
                     bssid = systemId + "_" + networkId + "_" + baseStationId;
                     type = NetworkType.CDMA;
@@ -1286,9 +1287,12 @@ public class WifiReceiver extends BroadcastReceiver {
         //ALIBI: MCC is always 3 chars, MNC may be 2 or 3.
         final String mnc = operatorCode.substring(3, operatorCode.length());
         final String mcc = operatorCode.substring(0, 3);
-        //DEBUG:
-        MainActivity.info("Operator MCC: "+mcc+" MNC: "+mnc);
-        return OPERATOR_MAP.get(mcc).get(mnc).getOperator();
+        //DEBUG:  MainActivity.info("Operator MCC: "+mcc+" MNC: "+mnc);
+        Map<String, MccMncRecord> country  = OPERATOR_MAP.get(mcc);
+        if (null != country) {
+            return country.get(mnc).getOperator();
+        }
+        return null;
     }
 
     private Map<String,Map<String,MccMncRecord>> initOperatorMap(AssetManager assetManager) {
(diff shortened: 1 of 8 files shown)


wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/listener/WifiReceiver.java
```diff
@@ -859,4 +1014,324 @@ private boolean doWifiScan() {
...
+     * @param operatorCode
+     * @return
+     */
+    private String getOperatorName(final String operatorCode) {
+        //ALIBI: have to work back from end, since MNC is always 3 chars, MCC may be 2 or 3.
+        final int boundary = operatorCode.length() -3;
+        final String mnc = operatorCode.substring(boundary, operatorCode.length());
+        final String mcc = operatorCode.substring(0, boundary);
+        //DEBUG: MainActivity.info("Operator MCC: "+mcc+" MNC: "+mnc);
+        return OPERATOR_MAP.get(mcc).get(mnc).getOperator();
```
uhtu: what happens when these .get() calls miss?

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/listener/WifiReceiver.java
```diff
@@ -859,4 +1014,324 @@ private boolean doWifiScan() {
...
+            stream.close();
+            ObjectMapper mapper = new ObjectMapper();
+            json = new String(buffer, "UTF-8");
+            MainActivity.info("operator load complete.");
+            JSONObject jObject = new JSONObject(json);
+            Iterator<String> mccKeys = jObject.keys();
+            while( mccKeys.hasNext() ){
+                String key = (String)mccKeys.next();
+                JSONObject country = jObject.getJSONObject(key);
+                Map<String, MccMncRecord> mccMap = null;
```
uhtu: consider just doing this like

Map<String, MccMncRecord> mccMap = map.get(key);
if (mccMap == null){
   mccMap = new HashMap<>();
   map.put(key, mccMap);
}

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/listener/WifiReceiver.java
```diff
@@ -480,153 +540,248 @@ public String getNetworkTypeName() {
...
+        String bssid = null;
+        NetworkType type = null;
+        Network network = null;
+
+        //noinspection StatementWithEmptyBody
+        if ( cellLocation == null ) {
+            // ignore
+        } else if ( cellLocation.getClass().getSimpleName().equals("CdmaCellLocation") ) {
+            try {
+                final int systemId = (Integer) cellLocation.getClass().getMethod("getSystemId").invoke(cellLocation);
```
uhtu: if this is a likely case, we should probably extract those accessors?

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/listener/WifiReceiver.java
```diff
@@ -859,4 +1014,324 @@ private boolean doWifiScan() {
...
+        Map<String,Map<String, MccMncRecord>> map = new HashMap<>();
+        String json = null;
+        try {
+            final InputStream stream = assetManager.open("mcc-mnc-dict.json");
+            MainActivity.info("operator stream: " + stream);
+
+            int size = stream.available();
+            byte[] buffer = new byte[size];
+
+            stream.read(buffer);
```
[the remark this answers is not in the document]
rksh: I was just punting to JSON deserialization to make us aware of failed reads, since we're unlikely to be able to detect the problem based on any non-zero read-bytes number.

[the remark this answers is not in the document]
uhtu: not even problem detection, that's just bog-standard stream mechanics, no?

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/listener/WifiReceiver.java
rksh: gonna punt this one to @bobzilladev - not sure why we did this!

rksh: aha, old trick to check version compat. all three calls are safe since SDK v. 5.

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/listener/WifiReceiver.java
```diff
@@ -1286,9 +1287,12 @@ private String getOperatorName(final String operatorCode) {
         //ALIBI: MCC is always 3 chars, MNC may be 2 or 3.
         final String mnc = operatorCode.substring(3, operatorCode.length());
         final String mcc = operatorCode.substring(0, 3);
-        //DEBUG:
-        MainActivity.info("Operator MCC: "+mcc+" MNC: "+mnc);
-        return OPERATOR_MAP.get(mcc).get(mnc).getOperator();
+        //DEBUG:  MainActivity.info("Operator MCC: "+mcc+" MNC: "+mnc);
+        Map<String, MccMncRecord> country  = OPERATOR_MAP.get(mcc);
+        if (null != country) {
+            return country.get(mnc).getOperator();
```
uhtu: can that second .get()  also nre? elvis operator wru?.

bobzilladev:
this is pretty epic. maybe turn off the heavy logging by default with an easy to flip bool.

PR #236, opened 2018-07-09
ENH - Cell enhancements

rksh:
first pass as DB-based version of this instead of in-memory map.
This begs a few interesting questions:
1. is *FCN the right substitute for channel, or would on-client translation to band be more useful?
2. will the server side need to be updated to respect network type + band/*FCN?
3. how should we trigger upgrades of the database on device when the MCC/MNC db updates remotely?
4. What improvements to client-side display should we make?
5. Should we version CSV format for this?


bobzilladev:
1. Is *FCN the right substitute for channel, or would on-client translation to band be more useful?
Don't know enough to comment. As long as it's something future proof. An integer for channel is best.
1. Will the server side need to be updated to respect network type + band/*FCN?
Server ignores channel for cell's currently, so we can implement whatever we want here.
1. How should we trigger upgrades of the database on device when the MCC/MNC db updates remotely?
This is a separate DB file for static data, I would say we just always copy the new one over the old one. Maybe need a version number built in to check?
1. What improvements to client-side display should we make?
I'd like to see the mcc mnc carrier displayed where we show oui manufacturer in the list and instance views.
1. Should we version CSV format for this?
This is backwards compatible, I don't think that's necessary. We have the client's version number already if we ever do need to differentiate.

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/NetworkActivity.java
```diff
@@ -134,18 +137,59 @@ public void onCreate(Bundle savedInstanceState) {
...
+                tv.setText( getString(R.string.na) );
+            } else {
+                final String[] cellCapabilities = network.getCapabilities().split(";");
+                tv.setText(cellCapabilities[0]+" "+channelCodeTypeForNetworkType(network.getType())+" "+chan);
             }
 
             tv = (TextView) findViewById( R.id.na_cap );
             tv.setText( " " + network.getCapabilities().replace("][", "]  [") );
 
+            if ( ! NetworkType.WIFI.equals(network.getType()) && ! NetworkType.CDMA.equals(network.getType())) {
```
bobzilladev: Seems like this should specifically check for GSM, rather than not other stuff, so bluetooth wouldn't trigger this.

rksh: this is shorthand for LTE, GSM, and WCDMA now - but point taken!

PR #239, opened 2018-07-14
FIX - fixing signal strength filter for MAX_VALUE

rksh:
new cell logging allowed Integer.MAX_VALUE into signal strengths.


bobzilladev:
i'd pry set a normalized as a new variable to not have the code duplicated, but that's just me :)

PR #242, opened 2018-07-22
ENH- Brootoof v0

rksh:
There a lot here, and a lot left to be done before this is production ready, but I'd like to get an initial review. General notes:
1. bluetooth (standard) scanning is keyed off of WifiReceiver's scan, but checks to make sure the previous discovery period has ended before starting a new one. In practice on an MotoZ Play, this means ~ every 20s.
2. bluetooth LE does a discovery cycle off of the same clock, but batches returns rather than receiving individual callbacks on each network.
3. there's currently an english text dictionary to double check reported device classes against, but there's a sizable icon library for known types present at several resolutions to help with i18n (intention is that the words can go away once we're satisfied.) no icon shown for Misc and Uncategorized BT nets.
4. We're overriding the frequency field *even more* than with cells - since the networkClass is an int, we're shoving it in there. This is probably a hack we shouldn't be doing.


wiglewifiwardriving/build.gradle
```diff
@@ -63,6 +63,7 @@ dependencies {
     implementation 'com.fasterxml.jackson.core:jackson-annotations:2.9.0.pr3'
     implementation 'com.fasterxml.jackson.core:jackson-databind:2.9.0.pr3'
     implementation 'ru.egslava:MaskedEditText:1.0.5'
+    implementation 'uk.co.alt236:bluetooth-le-library-android:1.0.0'
```
rksh: these guys are up to 1.1.1 but haven't published it to maven yet

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/ListFragment.java
```diff
@@ -92,6 +93,8 @@
     public static final String PREF_BATTERY_KILL_PERCENT = "batteryKillPercent";
     public static final String PREF_MUTED = "muted";
     public static final String PREF_WIFI_WAS_OFF = "wifiWasOff";
+    public static final String PREF_BT_WAS_OFF = "btWasOff";
+    public static final String PREF_SCAN_BT = "scanBluetooth";
```
rksh: making this settings-configurable for now, as it will be a beta feature.

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/ListFragment.java
```diff
@@ -209,9 +215,34 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container, Bundle sa
         MainActivity.info("setupLocation");
         setupLocation(view);
 
+        //TODO: if we want to hide BT when not scanning.
+        //final SharedPreferences prefs = getActivity().getSharedPreferences(SHARED_PREFS, 0);
+        //if (!prefs.getBoolean(ListFragment.PREF_SCAN_BT, false)) {
+        //    toggleBluetoothStats(false);
+        //}
+
```
rksh: i kinda like having the "B:0" as a hint that the functionality exists on the list view, but I'm convinceable.

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/ListFragment.java
```diff
@@ -228,11 +259,21 @@ private void setNetCountUI( final State state, final View view ) {
             return;
         }
         TextView tv = (TextView) view.findViewById( R.id.stats_run );
-        tv.setText( getString(R.string.run) + ": " + state.wifiReceiver.getRunNetworkCount() );
-        tv = (TextView) view.findViewById( R.id.stats_new );
-        tv.setText( getString(R.string.new_word) + ": " + state.dbHelper.getNewNetworkCount() );
+        long netCount = state.wifiReceiver.getRunNetworkCount();
```
rksh: Q: we could just modify getNetworkCount to perform this check everywhere (does the lockscreen/status badge need updating?)

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/ListFragment.java
```diff
@@ -228,11 +259,21 @@ private void setNetCountUI( final State state, final View view ) {
...
+        if (null != state.bluetoothReceiver){
+            netCount += state.bluetoothReceiver.getRunNetworkCount();
+        }
+        tv.setText( getString(R.string.run) + ": " + netCount );
+        //tv = (TextView) view.findViewById( R.id.stats_new );
+        //tv.setText( getString( R.string.new_word) +": " );
```
rksh: these would add a new label, but that gets impractical real fast, spacing-wise. perhaps is we shared "Cells" layout column with it?

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/listener/BluetoothReceiver.java
```diff
@@ -0,0 +1,555 @@
...
+            }
+        }
+
+        if (Build.VERSION.SDK_INT >= 21) {
+            final BluetoothLeScanner bluetoothLeScanner = bluetoothAdapter.getBluetoothLeScanner();
+            if (bluetoothLeScanner == null) {
+                MainActivity.info("bluetoothLeScanner is null");
+            }  else {
+                if (scanning.compareAndSet(false, true)) {
+                    final ScanSettings.Builder scanSettingsBulder = new ScanSettings.Builder();
```
rksh: weird typo

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/FilterActivity.java
```diff
@@ -71,6 +71,8 @@ public void onTextChanged( final String s ) {
                 ListFragment.FILTER_PREF_PREFIX + ListFragment.PREF_MAPF_CELL, true );
         MainActivity.prefBackedCheckBox( this, view, R.id.enabled,
                 ListFragment.FILTER_PREF_PREFIX + ListFragment.PREF_MAPF_ENABLED, true );
+        MainActivity.prefBackedCheckBox(this, view, R.id.showbt,
+                ListFragment.FILTER_PREF_PREFIX + ListFragment.PREF_MAPF_BT, true);
```
bobzilladev: Wondering if we should default off. But what the heck, might as well have on.

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/ListFragment.java
bobzilladev: fine with leaving it there at all times.

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/ListFragment.java
bobzilladev: fine with me. or make a utility function everyone can call.

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/ListFragment.java
bobzilladev: Unknown on this one, spacing is tight.

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/FilterActivity.java
rksh: i figure scanning is off by default for now - display is light by comparison

PR #245, opened 2018-08-01
FIX - fixing location race condition for bt scans.

rksh:
Doing some cleanup based on previous comments
Fixing a location race condition


wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/listener/BluetoothReceiver.java
```diff
@@ -312,13 +319,13 @@ public void bluetoothScan() {
                 MainActivity.info("bluetoothLeScanner is null");
             }  else {
                 if (scanning.compareAndSet(false, true)) {
-                    final ScanSettings.Builder scanSettingsBulder = new ScanSettings.Builder();
-                    scanSettingsBulder.setScanMode(ScanSettings.SCAN_MODE_LOW_POWER);
+                    final ScanSettings.Builder scanSettingsBuilder = new ScanSettings.Builder();
+                    scanSettingsBuilder.setScanMode(ScanSettings.SCAN_MODE_LOW_POWER);
```
bobzilladev: bulderdash!

PR #248, opened 2018-08-05
ENH - BLE devicetype

rksh:
follow on PR to #247  to max/match data from BT/BLE scans.


rksh:
probably better to make a way to update NetworkType in-place. working on that.

PR #265, opened 2018-08-24
FIX - one way to mitigate DB unreliability.

rksh:
don't want new cell info to reduce reliability. this isn't elegant, but it's unclear what else we can do.


wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/listener/WifiReceiver.java
```diff
@@ -992,24 +997,28 @@ private Network addOrUpdateCell(final String bssid, final String operator,
...
+                final LatLng LatLng = new LatLng( location.getLatitude(), location.getLongitude() );
+                network.setLatLng( LatLng );
+            }
 
-        if ( location != null ) {
-            dbHelper.addObservation(network, location, newForRun);
+            if ( location != null ) {
+                dbHelper.addObservation(network, location, newForRun);
+            }
+        } catch (SQLException sex) {
```
bobzilladev: hot

PR #266, opened 2018-08-27
FIX - a different approach for fallback

rksh:
since the DB errors are limited to 4.4.x samsung devices.


bobzilladev:
oh, of course... nice find

PR #269, opened 2018-08-27
FIX - stop trying to batch NetworkListView updates

rksh:
this makes the UI update/interaction a little wonky because of the high update frequency, but it should be safe.
Addressing 3.24 Beta errors in play console.


rksh:
https://stackoverflow.com/questions/3132021/android-listview-illegalstateexception-the-content-of-the-adapter-has-changed has some guidelines.

PR #271, opened 2018-09-02
ENH - one way of doing this. I don't know if this is perfectly safe yet

rksh:
flowing "batch" vs. "single net" down into the network add/update. This should be safer, but may not be enough without synchronizing access to the ListView Adapter.


bobzilladev:
let's aim for the stars!!!!

PR #273, opened 2018-10-15
ENH - okhttp upload

rksh:


wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/background/CountingRequestBody.java
```diff
@@ -0,0 +1,88 @@
+package net.wigle.wigleandroid.background;
```
bobzilladev: if this is just demo, should it go away before checkin? or is it handy?

bobzilladev: nevermind, firefox search didn't find it being used below.

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/background/CountingRequestBody.java
```diff
@@ -0,0 +1,88 @@
...
+import okhttp3.RequestBody;
+import okio.Buffer;
+import okio.BufferedSink;
+import okio.ForwardingSink;
+import okio.Okio;
+import okio.Sink;
+
+/**
+ * From https://github.com/hongyangAndroid/okhttputils/blob/master/okhttputils/src/main/java/com/zhy/http/okhttp/request/CountingRequestBody.java
+ * - Apache Licensed, here for demo only at this point
```
bobzilladev: Actually used now :)

rksh: updated - good catch

PR #274, opened 2018-10-16
fix some stacks seen in the wild

bobzilladev:


wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/ui/NetworkListAdapter.java
```diff
@@ -211,8 +211,14 @@ public  Network getItem(int pPosition) {
...
-        //should i just hash the object?
-        return networks.get(pPosition).getBssid().hashCode();
+        try {
+            //should i just hash the object?
+            return networks.get(pPosition).getBssid().hashCode();
+        }
+        catch (final IndexOutOfBoundsException ex) {
+            MainActivity.info("index out of bounds on getItem: " + pPosition + " ex: " + ex, ex);
+        }
+        return 0L;
```
rksh: this seems legit, but I'm worried the listview that updates from multiple sources is still a problem. probably needs a larger refactor.

bobzilladev: yeah, my logging didn't find multiple threads at work, but obviously there's a race somewhere here.

PR #280, opened 2018-11-04
Update german translation with minor fixes

designer2k2:
Hello,
with the latest update (bluetooth logging) some small translation errors where introduced.
I´m a native german speaker and this changes make more sense to read than the previous ones. 

greetings from Austria,
Stephan


bobzilladev:
Thanks!

PR #282, opened 2018-11-18
FIX - Bt fixes

rksh:
all kinds of bug reports about bluetooth. there's probably more to do, but these are batched up from travel, not good practice.

### DIFF
--- a/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/MainActivity.java
+++ b/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/MainActivity.java
@@ -1881,6 +1881,7 @@ public final class MainActivity extends AppCompatActivity {
         try {
             info("unregister bluetoothReceiver");
             unregisterReceiver( state.bluetoothReceiver );
+            state.bluetoothReceiver = null;
         } catch ( final IllegalArgumentException ex ) {
             //ALIBI: it's fine to call and fail here.
             info( "bluetoothReceiver not registered: " + ex );
--- a/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/MainActivity.java
+++ b/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/MainActivity.java
@@ -1853,20 +1853,18 @@ public final class MainActivity extends AppCompatActivity {
                 edit.putBoolean(ListFragment.PREF_BT_WAS_OFF, false);
             }
             edit.commit();
+            if ( state.bluetoothReceiver == null ) {
+                MainActivity.info( "new bluetoothReceiver");
+                // bluetooth scan listener
+                // this receiver is the main workhorse of bluetooth scanning
+                state.bluetoothReceiver = new BluetoothReceiver( this, state.dbHelper );
+                state.bluetoothReceiver.setupBluetoothTimer(true);
+            }
+            info("register bluetooth BroadcastReceiver");
+            final IntentFilter intentFilter = new IntentFilter(BluetoothDevice.ACTION_FOUND);
+            intentFilter.addAction(BluetoothAdapter.ACTION_DISCOVERY_FINISHED);
+            registerReceiver(state.bluetoothReceiver, intentFilter);
         }
-        if ( state.bluetoothReceiver == null ) {
-            MainActivity.info( "new bluetoothReceiver");
-            // bluetooth scan listener
-            // this receiver is the main workhorse of bluetooth scanning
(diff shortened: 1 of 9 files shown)


wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/MainActivity.java
```diff
@@ -1875,12 +1873,15 @@ public void endBluetooth(SharedPreferences prefs) {
         }
 
         final BluetoothAdapter bluetoothAdapter = BluetoothAdapter.getDefaultAdapter();
-        if (bluetoothAdapter != null && bluetoothAdapter.isDiscovering()) {
-            bluetoothAdapter.cancelDiscovery();
+        if (bluetoothAdapter != null) {
+            if (bluetoothAdapter.isDiscovering()) {
+                bluetoothAdapter.cancelDiscovery();
+            }
```
bobzilladev: This is a no-op change due to conditional short-circuiting, what was the goal here?

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/MappingFragment.java
```diff
@@ -188,8 +189,12 @@ public void onMapReady(final GoogleMap googleMap) {
                 googleMap.setBuildingsEnabled(true);
                 final boolean showTraffic = prefs.getBoolean(ListFragment.PREF_MAP_TRAFFIC, true);
                 googleMap.setTrafficEnabled(showTraffic);
-                final int mapType = prefs.getInt(ListFragment.PREF_MAP_TYPE, GoogleMap.MAP_TYPE_NORMAL);
-                googleMap.setMapType(mapType);
+                if (null != prefs) {
```
bobzilladev: Prefs can't be null here or it would have blown up on line 190.

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/listener/BluetoothReceiver.java
```diff
@@ -632,10 +654,14 @@ private Network addOrUpdateBt(final String bssid, final String ssid,
         final ConcurrentLinkedHashMap<String, Network> networkCache = MainActivity.getNetworkCache();
         final boolean showCurrent = prefs.getBoolean(ListFragment.PREF_SHOW_CURRENT, true);
 
-        final boolean newForRun = runNetworks.add(bssid);
+        final boolean newForRun = runNetworks.add(bssid) && networkCache.containsKey(bssid);
```
bobzilladev: Why should this be in the cache already to be new?

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/MappingFragment.java
rksh: your guess is as good as mine, but there it is in the crash report!

bobzilladev: Maybe cover the prefs above, and any others that might be missed?

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/ui/NetworkListAdapter.java
```diff
@@ -185,11 +191,20 @@ public void batchUpdateBt(final boolean showCurrent, final boolean updateLe, fin
         }
         if (updateLe) {
             leNets.addAll(nextLeNets);
-            networks.addAll(leNets);
+            for (Network leNet: leNets) {
+                if (!networks.contains(leNet)) {
```
bobzilladev: hrm, these are O(n) searches in a loop, seems expensive.

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/MappingFragment.java
rksh: done!

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/MainActivity.java
rksh: wanted a place to explicitly null out bluetoothAdapter reference below.

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/ui/NetworkListAdapter.java
rksh: See #284 for partitioning

PR #283, opened 2018-11-27
FIX - Mcx db fixes

rksh:
follow on to BT fixes to handle the cell lookup issues on certain phone models.

### DIFF
--- a/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/model/GsmOperator.java
+++ b/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/model/GsmOperator.java
@@ -1,6 +1,7 @@
 package net.wigle.wigleandroid.model;
 
 import android.annotation.TargetApi;
+import android.database.sqlite.SQLiteDatabaseCorruptException;
 import android.os.Build;
 import android.telephony.CellIdentityGsm;
 import android.telephony.CellIdentityLte;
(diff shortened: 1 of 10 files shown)


wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/model/GsmOperator.java
```diff
@@ -256,8 +257,20 @@ private static boolean recognizedMnc(final String mcc, final String mnc) {
...
-                operator = s.mxcDbHelper.networkNameForMccMnc(mcc, mnc);
+            if ((null != s) && (null != s.mxcDbHelper)) {
+                try {
+                    operator = s.mxcDbHelper.networkNameForMccMnc(mcc, mnc);
+                } catch (SQLiteDatabaseCorruptException sqldbex) {
+                    // this case seems isolated to LG android 4.4 devices
+                    MainActivity.warn("Mxc DB corrupt - unable to lookup carrier", sqldbex);
+                    try {
+                        //recopy the MccMnc DB file to see whether we can recover.
+                        s.mxcDbHelper.implantMxcDatabase();
```
bobzilladev: Maybe keep a counter to not attempt this more than n times?

rksh: hm, good idea.

PR #284, opened 2018-12-01
ENH - Set backed net list, self-contained

rksh:
Per feedback on #282 - a contingent PR that moves set-backed list into its own class.
PLEASE REVIEW HARD

### DIFF
--- a/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/ui/SetBackedNetworkList.java
+++ b/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/ui/SetBackedNetworkList.java
@@ -1,0 +1,495 @@
+package net.wigle.wigleandroid.ui;
+
+import android.support.annotation.NonNull;
+
+import net.wigle.wigleandroid.MainActivity;
+import net.wigle.wigleandroid.model.Network;
+
+import java.lang.reflect.Array;
+import java.util.AbstractList;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.List;
+import java.util.ListIterator;
+import java.util.Set;
+
+/**
+ * Because a SetList is for bands.
+ */
+public class SetBackedNetworkList extends AbstractList<Network> implements List<Network> {
+    private final List<Network> unsafeNetworks = new ArrayList<>();
+    private final List<Network> networks = Collections.synchronizedList(unsafeNetworks);
+
+    private final Set<Network> btNets = new HashSet<>();
+    private final Set<Network> leNets = new HashSet<>();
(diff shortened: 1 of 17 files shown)


bobzilladev:
seems cromulent

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/ui/SetBackedNetworkList.java
```diff
@@ -0,0 +1,495 @@
...
+        return succeeded && (btSuccess || leSuccess || wifiSuccess || cellSuccess);
+    }
+
+    @Override
+    public boolean retainAll(@NonNull Collection<?> collection) {
+        btNets.retainAll(collection);
+        leNets.retainAll(collection);
+        wifiNets.retainAll(collection);
+        cellNets.retainAll(collection);
+        Set retainNets = new HashSet<Network>();
```
uhtu: so, can this be a List<> since we already know the sets won't overlap?

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/ui/SetBackedNetworkList.java
```diff
@@ -0,0 +1,495 @@
...
+                offset++;
+            }
+        }
+        return false;
+    }
+
+    @Override
+    public boolean removeAll(@NonNull Collection<?> collection) {
+        //DEBUG (1/3): long size = networks.size();
+        final boolean succeeded = networks.removeAll(collection);
```
uhtu: would this work to just networks.clear() and then re-add the four sets after they've .removeAll() ?

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/ui/SetBackedNetworkList.java
rksh: so, the clear-and-repop was my initial implementation, but then we lose the the sort ordering that backs the display.

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/ui/SetBackedNetworkList.java
rksh: same case as above - we lose sorting.

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/ui/SetBackedNetworkList.java
rksh: perf testing indicates that an ArrayList in place of a Set is approximately a 2x performance decrease (?!) average over ten runs.

uhtu:
:+1:

Issue #227, opened 2018-06-02
Exclude "Scanning turned off" time from Run duration


goldfndr:
I think the Run duration amount would be better implemented as only incrementing when Scanning is turned on. Benefits from this:

* If you scan on the way to work, turn off scanning while at work, then resume scanning after work, it wouldn't look like the device was scanning on its own.
* Additional indication that scanning is turned off.
* Easier to roughly calculate average of how often you're seeing APs.

Perhaps a preference toggle between Run duration and Scan duration?

rksh:
good point

rksh:
in the next release!

PR #286, opened 2018-12-17
ENH - Timediff request

rksh:
fix for #227 - includes scan and run durations side-by-side


bobzilladev:
mmm, tartar sauce. +1

PR #291, opened 2018-12-23
[ENH] - Activity refactor

rksh:
kinda a long refactor. not advocating for merge, but demonstrates how this would works.
Still doesn't turn the last menu item red. :/

PR #297, opened 2019-01-02
ENH - Actionbar search stats

rksh:
1. making all three stats elements (user/site/rank) sub-elements of the master stats element in the menu, minimizing menu size for casual use/in-car safety
2. adding a "search" menu item that exposes both the local DB search as well as searching the online database for logged-in users.

This is a first cut, and will require refinement. This is a follow-on PR for https://github.com/wiglenet/wigle-wifi-wardriving/pull/291 which is a major change, and won't be merged until 2.39 beta feedback has been received and incorporated.


rksh:
merged into beta.

PR #299, opened 2019-01-07
[ENH] - M8b export

rksh:
a Proof-of-concept m8b ( https://github.com/wiglenet/m8b ) output option for your stumbling DBs.
still requires i18n, extensive review.

Is this the right way to do this?

NB: duplicate intermediate filepath calculation, hard to do correctly because of signatures in anon classes.


rksh:
replacing with PR vs. beta.

PR #301, opened 2019-01-10
FIX - Console error cleanup

rksh:
ditching a spurious log message and trying to fix a production null problem.


rksh:
sorry, a couple of items that got left out of the initial PR.

bobzilladev:
lgtm. interesting, we might be forced to bump our min off of 14 at some point.

PR #302, opened 2019-01-11
[FIX] - trying to do DB lookups under 4.4 pre 4.4.1 explodes a lot

rksh:
stability for older platforms.


wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/db/MxcDatabaseHelper.java
```diff
@@ -107,6 +107,11 @@ public void onUpgrade(SQLiteDatabase sqLiteDatabase, int i, int i1) {
     public MccMncRecord networkRecordForMccMnc(final String mcc, final String mnc) throws SQLException {
         Cursor cursor = null;
 
+        // ALIBI: old, incompatible DB implementation
+        if (android.os.Build.VERSION.SDK_INT == 19) {
```
bobzilladev: should that be <= 19?

PR #321, opened 2019-03-19
[FIX] - Safe mcc mnc handling

rksh:
maybe addressing some infrequent weird behavior we see by improving SQLite DB handling.


bobzilladev:
diff looks fine. the right way is pry try-with-resources at some point.

PR #326, opened 2019-03-30
add bluetooth stats everywhere

bobzilladev:


bobzilladev:
i made a new line for uploads, and others were already as lists, or replaced a duplicate wifi with bt, so i don't think there'll be layout issues.

PR #332, opened 2019-05-02
[FIX] - fixing a minor UI big with quickbutton for map.

rksh:
navigation drawer selection could get out of sync.

### DIFF
--- a/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/ListFragment.java
+++ b/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/ListFragment.java
@@ -401,18 +401,14 @@ public final class ListFragment extends Fragment implements ApiListener, DialogL
         MenuItem item = menu.add(0, MENU_MAP, 0, getString(R.string.tab_map));
         item.setIcon( android.R.drawable.ic_menu_mapmode );
         item.setShowAsAction(MenuItem.SHOW_AS_ACTION_IF_ROOM);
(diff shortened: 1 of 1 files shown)


wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/ListFragment.java
```diff
@@ -397,16 +400,19 @@ public void onDetach() {
...
         MenuItem item = menu.add(0, MENU_MAP, 0, getString(R.string.tab_map));
         item.setIcon( android.R.drawable.ic_menu_mapmode );
-        MenuItemCompat.setShowAsAction(item, MenuItemCompat.SHOW_AS_ACTION_IF_ROOM);
+        item.setShowAsAction(MenuItem.SHOW_AS_ACTION_IF_ROOM);
+        //MenuItemCompat.setShowAsAction(item, MenuItemCompat.SHOW_AS_ACTION_IF_ROOM);
 
         item = menu.add(0, MENU_FILTER, 0, getString(R.string.menu_ssid_filter));
         item.setIcon(android.R.drawable.ic_menu_search);
-        item.setIcon(android.R.drawable.ic_menu_manage);
-        MenuItemCompat.setShowAsAction(item, MenuItemCompat.SHOW_AS_ACTION_IF_ROOM);
```
bobzilladev: it looks like the icon is changing here, from manage to search, is that desired? otherwise lgtm.

rksh: a reasonable question; i noticed redundancy, but i removed the effective one, i think. I'll swap.

PR #335, opened 2019-05-08
[ENH] here's a naive way of moving lerp into the GPS listener

rksh:
not sure about all the consequences of doing this, but it looks safe. review would be greatly appreciated.


bobzilladev:
The feels like a better place for it. I have no idea the consequences.

PR #339, opened 2019-06-01
use gnns status if available

bobzilladev:


wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/MainActivity.java
```diff
@@ -2126,6 +2153,15 @@ private void internalSetLocationUpdates(final long updateIntervalMillis, final f
         }
     }
 
+    public static <A> String join(final String delimiter, final Iterable<A> iterable) {
```
rksh: i can't fault you for doing this. android is slow. my only caution is null checking.

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/listener/GPSListener.java
```diff
@@ -358,16 +371,61 @@ private boolean horribleGps(final Location location) {
...
         }
         return satCount;
     }
 
+    public Map<String, Integer> getConstellations() {
+        final Map<String, Integer> cons = new TreeMap<>();
+        if (gnssStatus != null && Build.VERSION.SDK_INT >= 24) {
+            for ( int i = 0; i < gnssStatus.getSatelliteCount(); i++ ) {
+                if ( gnssStatus.usedInFix(i) ) {
+                    final String key = constellationToString(gnssStatus.getConstellationType(i));
```
rksh: what do these look like x-lang?

bobzilladev: It's a static proper noun, so same in all languages.

bobzilladev:
Pushed null checks for join, and a fix to make it buildable in non-q-land.

PR #341, opened 2019-06-02
cap lerp at 200m

bobzilladev:


bobzilladev:
given the miles-long lines all over continents i'm happy to not pull in useless data.

PR #344, opened 2019-06-03
no longer search to find qr code

bobzilladev:


bobzilladev:
I think folks know, just like upc codes, they show up everywhere now a days.

PR #346, opened 2019-06-09
[FIX] micropr - shouldn't crash.

rksh:
just catching security exception if users say no to bt.


wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/MainActivity.java
```diff
@@ -1827,13 +1827,17 @@ private void setupWifiReceiverIntent() {
...
+        try {
+            final BluetoothAdapter bt = BluetoothAdapter.getDefaultAdapter();
+            if (bt == null) {
+                info("No bluetooth adapter");
+                return false;
+            }
+            if (!bt.isEnabled()) {
+                return true;
+            }
+        } catch (java.lang.SecurityException sex) {
```
bobzilladev: 🌶

PR #351, opened 2019-07-24
[FIX] Random stability experiments

rksh:
no proof that this will fix our current stability issues, but some possible culprits addressed


wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/listener/StartWigleAtBootReciever.java
```diff
@@ -12,6 +12,9 @@
 
     @Override
     public void onReceive(Context context, Intent intent) {
+        if (null == intent) {
+            return;
```
bobzilladev: No log message on this one, odd one out.

bobzilladev: He didn't want to log here since things aren't all running yet.

PR #357, opened 2019-08-06
[FIX?] seeing whether deregistration of the GNSS listener fixes our RSEs

rksh:
Not sure that this is the cause yet, but it fits the pathology:
- stochastic background process
- broadcast listener
- introduced in 2.44
- can happen without obvious external cause


rksh:
12 hours and counting, no RSE...

PR #364, opened 2019-08-22
[ENH] warn when we can't implant mxc db, or if storage is almost full.

rksh:
- seeing what's left of our SQL problems if we tell people not to stumble with full devices.
- language pack updates... the endless struggle.

### DIFF
--- a/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/MainActivity.java
+++ b/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/MainActivity.java
@@ -83,7 +83,9 @@ import net.wigle.wigleandroid.model.ConcurrentLinkedHashMap;
 import net.wigle.wigleandroid.model.Network;
 import net.wigle.wigleandroid.ui.SetNetworkListAdapter;
 import net.wigle.wigleandroid.ui.WiGLEToast;
+import net.wigle.wigleandroid.util.FileUtility;
 import net.wigle.wigleandroid.util.InstallUtility;
+import net.wigle.wigleandroid.util.InsufficientSpaceException;
 
 import java.io.File;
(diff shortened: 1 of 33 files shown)


wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/MainActivity.java
```diff
@@ -356,8 +360,24 @@ public void onCreate(Bundle savedInstanceState) {
         //if (!state.mxcDbHelper.isPresent()) {
         try {
             state.mxcDbHelper.implantMxcDatabase();
+        } catch (InsufficientSpaceException isex) {
```
bobzilladev: hot

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/MainActivity.java
```diff
@@ -2499,4 +2519,29 @@ public void backgroundUploadFile(){
             MainActivity.warn("Authentication failure on background run upload");
         }
     }
+
+    public boolean checkStorage() {
+        boolean safe;
+        boolean external = hasSD();
```
bobzilladev: Is this only called after the initial setup where we ask for permission for the external storage? Want to make sure it's happy on initial install.

rksh: oh good point.

rksh: looks like `onCreate` calls `setupPermissions` almost immediately - you might still get the warning if you haven't agreed and local storage is really tight, but on "accept", you'd relaunch with the new perms, perform the check again, and get `hasSD() == true`

PR #368, opened 2019-08-31
fix npe in maprender

bobzilladev:


rksh:
don't know why this is happening, but I feel like we're doing a lot of dynamic data mgmt now, and we're still finding the edges of it, with the rapid net list updates.

PR #370, opened 2019-08-31
Changes required for android Q (10)

bobzilladev:
The Time is Now.

### DIFF
--- a/wiglewifiwardriving/src/main/AndroidManifest.xml
+++ b/wiglewifiwardriving/src/main/AndroidManifest.xml
@@ -249,6 +249,7 @@
         <service
             android:name=".WigleService"
             android:enabled="true"
+            android:foregroundServiceType="location"
             android:icon="@mipmap/ic_launcher" />
         <uses-library
             android:name="com.google.android.maps"
--- a/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/MainActivity.java
+++ b/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/MainActivity.java
@@ -1954,8 +1954,12 @@ public final class MainActivity extends AppCompatActivity {
                 info("wifi turned on");
             }
             else {
-                info("wifi NOT turned on");
+                info("wifi NOT turned on, resultCode: " + resultCode);
             }
+        }
+        else {
+            info("MainActivity: Unhandled requestCode: " + requestCode
+                    + " resultCode: " + resultCode);
         }
     }
 
--- a/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/MainActivity.java
+++ b/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/MainActivity.java
@@ -175,6 +175,7 @@ public final class MainActivity extends AppCompatActivity {
     private static final String LOG_TAG = "wigle";
     public static final String ENCODING = "ISO-8859-1";
     private static final int PERMISSIONS_REQUEST = 1;
+    private static final int ACTION_WIFI_CODE = 2;
 
     static final String ERROR_STACK_FILENAME = "errorstack";
     static final String ERROR_REPORT_DO_EMAIL = "doEmail";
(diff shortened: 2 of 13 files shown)


wiglewifiwardriving/src/main/AndroidManifest.xml
```diff
@@ -249,6 +249,7 @@
         <service
             android:name=".WigleService"
             android:enabled="true"
+            android:foregroundServiceType="location"
```
rksh: Would we be future-proofing by adding ACCESS_BACKGROUND_LOCATION at this point as well (looks like FINE/COARSE imply this IFF we target < SDK 29 according to https://developer.android.com/preview/privacy/device-location )

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/MainActivity.java
```diff
@@ -406,6 +421,14 @@ private void pieScanSet(final SharedPreferences prefs, final String key) {
         }
     }
 
+    @SuppressLint("ApplySharedPref")
+    private void qScanSet(final SharedPreferences prefs, final String key) {
+        if (SCAN_P_DEFAULT == prefs.getLong(key, -1)) {
+            info("Removing 30 second scan for " + key + " due to less broken Android Q");
+            prefs.edit().remove(key).commit();
```
rksh: huzzah!

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/MainActivity.java
```diff
@@ -1910,14 +1937,28 @@ private void setupWifi() {
         // register wifi receiver
         setupWifiReceiverIntent();
 
-        if (state.wifiLock == null) {
+        if (state.wifiLock == null && wifiManager != null) {
             MainActivity.info("lock wifi radio on");
-            // lock the radio on
+            // lock the radio on (only works in 28 (P) and lower)
```
rksh: is WIFI_MODE_FULL_HIGH_PERF the new best-we-can-do? https://developer.android.com/reference/android/net/wifi/WifiManager.html#WIFI_MODE_FULL_HIGH_PERF

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/MainActivity.java
```diff
@@ -1910,14 +1937,28 @@ private void setupWifi() {
...
+    @Override
+    protected void onActivityResult(int requestCode, int resultCode, Intent data) {
+        // Check which request we're responding to
+        if (requestCode == ACTION_WIFI_CODE) {
+            // Make sure the request was successful
+            if (resultCode == RESULT_OK) {
+                info("wifi turned on");
+            }
+            else {
+                info("wifi NOT turned on");
```
rksh: should we log what the result code here so we know what other junk we're getting in response?

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/listener/GPSListener.java
```diff
@@ -418,7 +421,7 @@ private String constellationToString(final int constellationType) {
                 con = "Galileo";
                 break;
         }
-        if (Build.VERSION.SDK_INT > 28 && constellationType == 7 /*GnssStatus.CONSTELLATION_IRNSS*/) {
+        if (Build.VERSION.SDK_INT > 28 && constellationType == GnssStatus.CONSTELLATION_IRNSS) {
```
rksh: good memory

wiglewifiwardriving/src/main/AndroidManifest.xml
bobzilladev: Nope, that's an extra permission we won't ever need since we are foreground. Background location is limited to very few updates, I don't think it'd be useful, and it makes us a target for thinking we're bad actors.

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/MainActivity.java
bobzilladev: All the non-deprecated locks are not useful to us since they only matter when connected to an AP. "The lock is only active when the device is connected to an access point"

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/MainActivity.java
bobzilladev: updated

PR #372, opened 2019-09-10
[ENH] Despaz bluetooth display

rksh:
not perfect, but it makes the display less unreliable. prep for route logging/display with density.
also contains a small feature-detection fix.

### DIFF
--- a/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/listener/BluetoothReceiver.java
+++ b/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/listener/BluetoothReceiver.java
@@ -1,6 +1,5 @@
 package net.wigle.wigleandroid.listener;
 
-import android.app.ListActivity;
 import android.bluetooth.BluetoothAdapter;
 import android.bluetooth.BluetoothClass;
 import android.bluetooth.BluetoothDevice;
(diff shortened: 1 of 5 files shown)


wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/listener/BluetoothReceiver.java
```diff
@@ -140,13 +147,19 @@
     private long adUuidNoScanUuid = 0;
     private long scanUuidNoAdUuid = 0;
 
+    // prev/current sets of BSSIDs for each scan. ~ redundant w/ sets in SetBackedNetworkList in current-only mode...
+    private Set<String> latestBt = new HashSet<>();
+    private Set<String> prevBt = new HashSet<>();
 
-    public BluetoothReceiver(final MainActivity mainActivity, final DatabaseHelper dbHelper ) {
+    private Set<String> latestBtle = new HashSet<>();
+    private Set<String> prevBtle = new HashSet<>();
```
bobzilladev: are we sure these will only be touched by one thread?

rksh: I'm not perfectly sure the bt le scanner will always write from the same thread. the .size() calls could result in some concurrency risk. I'll change the implementation.

rksh: should be safe now

bobzilladev: the i* vars aren't used elsewhere, I'd just move the new hashset into the next line so no one can get a handle on the non-synchronized reference.

PR #373, opened 2019-09-30
[ENH] Route visualization and export

rksh:
there's a lot here; consider this a draft.
- new route table
- allows per-run route logging
- allows visualization of current run route on map
- allows export of current route (for now)
- removed some annoying log messages

### DIFF
--- a/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/DataFragment.java
+++ b/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/DataFragment.java
@@ -22,6 +22,8 @@ import android.app.ProgressDialog;
 import android.content.DialogInterface;
 import android.content.Intent;
 import android.content.SharedPreferences;
+import android.content.pm.PackageInfo;
+import android.content.pm.PackageManager;
 import android.database.Cursor;
 import android.media.AudioManager;
 import android.net.Uri;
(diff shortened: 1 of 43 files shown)


wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/db/DatabaseHelper.java
```diff
@@ -562,7 +612,10 @@ public void close() {
                     if ( updateNetworkMetadata != null ) {
                         updateNetworkMetadata.close();
                     }
-                    if ( db.isOpen() ) {
+                    if ( insertRoute != null ) {
+                        insertRoute.close();
+                    }
+                if ( db.isOpen() ) {
```
bobzilladev: nit: Spacing looks off

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/DataFragment.java
```diff
@@ -82,6 +86,12 @@
     private static final String M8B_EXTENSION = ".m8b";
     private static final int SLICE_BITS = 30;
 
+    private static final String GPX_DIR = "/wiglewifi/gpx/";
+    private static final String GPX_EXTENSION = ".gpx";
+    private final static String GPX_HEADER = "<?xml version=\"1.0\" encoding=\"UTF-8\" standalone=\"no\" ?><gpx xmlns=\"http://www.topografix.com/GPX/1/1\" creator=\"MapSource 6.15.5\" version=\"1.1\" xmlns:xsi=\"http://www.w3.org/2001/XMLSchema-instance\"  xsi:schemaLocation=\"http://www.topografix.com/GPX/1/1 http://www.topografix.com/GPX/1/1/gpx.xsd\"><trk>\n";
```
goldfndr: `creator=\"MapSource 6.15.5\"` ?

rksh: copypasta from back when this was a garmin standard! I've religiously included it like some kind of "quirks-mode" flag in various projects... my only concern with a more accurate answer here is that I don't know what it might break.

rksh: Added an honest creator tag - we'll see what breaks!

PR #374, opened 2019-09-30
[FIX] Freespace fix

rksh:
actually just masking an exception, but it's a very rare case, indicates some highly unusual storage configuration. first do no harm.


rksh:
bob's right - ugly.

PR #383, opened 2019-11-29
[ENH] removing compromised lib that we suspected of being a crash source

rksh:
uses a replacement animated `<vector>` item in place of the old GIF resource.

at original implementation, this feature was SDK 21 and up - but with AnimatedVectorDrawableCompat this goes all the way back to SDK 14, so we're within our support window!

Animation is different from the original because android animation of clipping paths doesn't appear compatible with our original design.

This will require EXTENSIVE version testing to make sure it plays nice with everything.


wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/ListFragment.java
```diff
@@ -296,14 +297,27 @@ public void setStatusUI( String status ) {
...
+            AnimatedVectorDrawableCompat scanningAnimation;
+            if (null != getActivity() && !animating) {
+                animating = true;
+                scanningAnimation = AnimatedVectorDrawableCompat.create(getActivity().getApplicationContext(), R.drawable.animated_wifi_simplified);
+                scanningImageButton.setImageDrawable(scanningAnimation);
+                scanningImageButton.setScaleType(ImageView.ScaleType.FIT_CENTER);
+                if (null != scanningAnimation ) {
+                    scanningAnimation.start();
+                }
+            } else {
```
bobzilladev: Could get here with a non-null activity, but true animating, pry need to put the !animating above in a sub-if.

bobzilladev: The error message says "Null activity context" but you can get here if activity is NOT null, but animating is true.
```if (null != getActivity() && !animating) {```

Comment (bobzilladev) [reply wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/ListFragment.java]: So you probably don't want a bunch of errors if animating is true, so maybe want:
```
if (null != getActivity()) {
  if (!animating) {
    ...
  }
} else {
  MainActivity.error("Null activity context - can't set animation");
}
```

PR #385, opened 2019-12-04
[ENH] User Stats view condensation/refactor

rksh:
as a frequent viewer of the user stats page, it was hard to read at a glance.

- updated "discovered" and "seen" entries for WiFi/BT/Cell into organized horizontal bars
- used WiGLE icons and style colors to help improve readability
- text heavy format was bad for responsive layout/translation
- many translations were absent for the existing text-heavy presentation

future directions:

- observations should be a section as well, but we don't condense cell and BT in the same way as wifi, so this is a half-measure until we add those stats.
- this/last month should be a section, but the API doesn't publish those yet

### DIFF
--- a/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/UserStatsFragment.java
+++ b/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/UserStatsFragment.java
@@ -1,6 +1,7 @@
 package net.wigle.wigleandroid;
 
 import android.annotation.SuppressLint;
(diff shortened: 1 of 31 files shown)


rksh:
Screenshot_20191203-190828
(BEFORE)

Screenshot_20191203-190643
(AFTER)

goldfndr:
On the website, there's no hyphen between the down arrow and 102 for the Month Ranking.

rksh:
re: hyphen - i believe that's the "sign" of the number!

PR #387, opened 2019-12-08
error stacks and mnc db can use internal filesystem

bobzilladev:


rksh:
basically everywhere we offer to write anything out, we'll need to audit. :/

PR #391, opened 2019-12-17
[ENH] Centralize FileSystem access - refactor refactor in preparation for Android R

rksh:
All requests involving paths now moved into FileUtility.
DRY'd up SD pathing, error messaging.
Tested in local and external FS mode on Android Pie, for
  - Upload KML View
  - Database Run KML
  - Database DB KML
  - Database M8B
  - Database CSV RUN
  - Database CSV DB
  - Database GPX
This is a huge change, and deserves extensive testing.

### DIFF
--- a/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/DataFragment.java
+++ b/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/DataFragment.java
@@ -61,9 +61,7 @@ import java.util.Set;
 import java.util.TreeMap;
 
 import static net.wigle.wigleandroid.ListFragment.PREF_LOG_ROUTES;
-import static net.wigle.wigleandroid.util.FileUtility.GPX_DIR;
 import static net.wigle.wigleandroid.util.FileUtility.GPX_EXT;
-import static net.wigle.wigleandroid.util.FileUtility.M8B_DIR;
 import static net.wigle.wigleandroid.util.FileUtility.M8B_EXT;
 import static net.wigle.wigleandroid.util.FileUtility.M8B_FILE_PREFIX;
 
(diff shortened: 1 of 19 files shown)


wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/DataFragment.java
```diff
@@ -769,28 +770,34 @@ protected String doInBackground(Integer... dbKRecords) {
...
+                    m8bDestFile = new File(a.getApplication().getFilesDir(),
                             M8B_FILE_PREFIX + M8B_EXT);
                 }
                 //ALIBI: always start fresh
-                boolean append = false;
-                out = new FileOutputStream(m8bDestFile, append).getChannel();
+                out = new FileOutputStream(m8bDestFile, false).getChannel();
 
             } catch (IOException ioex) {
-                MainActivity.error("Unable to open ouput: ", ioex);
```
uhtu: muh ouput!

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/background/BackgroundGuiHandler.java
```diff
@@ -143,63 +146,92 @@ public void handleMessage( final Message msg ) {
...
+                                    showError(fm, msg, status);
+                                    MainActivity.error("Null filePath - unable to share.");
+                                }
+                            } else {
+                                showError(fm, msg, status);
+                                MainActivity.error("null context - cannot generate intent to share KML");
+                            }
+                        } catch (Exception ex) {
+                            showError(fm, msg, status);
+                            MainActivity.error("Failed to send file intent: ", ex);
```
uhtu: how can you resist this being called "un intentional"?

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/DataFragment.java
```diff
@@ -163,64 +168,99 @@ public void transferComplete() {
...
 
     private void setupBackupDbButton( final View view ) {
-        final Button kmlExportButton = (Button) view.findViewById( R.id.backup_db_button );
-        if ( ! MainActivity.hasSD() ) {
-            kmlExportButton.setEnabled(false);
+        final Button dbBackupButton = view.findViewById( R.id.backup_db_button );
+
+        //TODO: there's a question here: if the backupDb option shouldn't be used w/out SD
+        //  should we also disable the export to KML/CSV options, since they'll certainly be larger?
+        if ( ! FileUtility.hasSD() ) {
```
bobzilladev: I just figured it was pointless to copy internally since it would be lost if the app was uninstalled. But if we can save internally and then provide options to share it from there to the other-apps zone, or via email/chat/etc options that you have, then it's still useful.

rksh: oh yeah, duh... i didn't think that through all the way! i guess we could just offer the share intent immediately, for people to use for backup!

PR #392, opened 2019-12-20
[FIX] removing unused BTLE lib

rksh:
relic of an older era


rksh:
interfered w/ FOSS master

PR #399, opened 2020-01-25
allow internal db backup sharing

bobzilladev:
This...seems to work...after a while...android is fun?

### DIFF
--- a/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/DataFragment.java
+++ b/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/DataFragment.java
@@ -228,14 +228,6 @@ public final class DataFragment extends Fragment implements ApiListener, Transfe
 
     private void setupBackupDbButton( final View view ) {
         final Button dbBackupButton = view.findViewById( R.id.backup_db_button );
-
-        //TODO: experiments extending the provider to grant direct share access to the
-        //   databases/wiglewifi.sqlite file failing under Pie. If we can add a provider that can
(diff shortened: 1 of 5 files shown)


wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/DataFragment.java
```diff
@@ -595,6 +587,9 @@ protected void onPostExecute( Integer result ) {
                     if (null == c) {
                         MainActivity.error("null context in DB backup postExec");
                     } else {
+                        final File backupFile = new File(dbResult.getSecond());
+                        MainActivity.info("backupfile: " + backupFile.getAbsolutePath());
+                        MainActivity.info("exists: " + backupFile.exists() + " read: " + backupFile.canRead());
```
rksh: maybe condense to a single log line for readability?

wiglewifiwardriving/src/main/res/xml/sqlite_paths.xml
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
 <paths xmlns:android="http://schemas.android.com/apk/res/android">
     <external-path name="sqlite" path="wiglewifi"/>
-    <files-path name="sqlite-internal" path="./"/>
+    <files-path name="sqlite-internal" path="backups"/>
```
rksh: would "sqlite" be a more consistent naming convention for this dir? I can see both sides.

bobzilladev:
Updated

PR #400, opened 2020-01-29
[ENH] Uploads status/download/CSV sharing

rksh:
This is a set of workarounds for the new Android 10 and up file semantics.
1. displays local, remote data presence in an easy-to-read widget (I hope)
2. If not local, allows download+storage+re-incorporation of the upload into the DB.
3. If present locally, allows "export" of CSV via share intent

**_known limitations:_**
- channel information not available from WiGLE type=WIFI networks currently
- probably can't download Kismet uploads
- Because download infra is string-based, downloads CSV over gzipped pipe, compresses to local GZ file (rather than downloading GZ from server)
- not optimized
- i18n not done yet


rksh:
ok, I think this thing is ready to test. I've seen some weird edge cases where I think we're uploading downloaded points, but 1. that won't hurt anything 2. I can't reliably repro it / figure out why that would happen

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/db/DatabaseHelper.java
```diff
@@ -568,6 +588,9 @@ else if ( db.getVersion() == 1 ) {
         insertLocation = db.compileStatement( "INSERT INTO " + LOCATION_TABLE
                 + " (bssid,level,lat,lon,altitude,accuracy,time) VALUES (?,?,?,?,?,?,?)" );
 
+        insertLocationExternal = db.compileStatement( "INSERT INTO " + LOCATION_TABLE
+                + " (bssid,level,lat,lon,altitude,accuracy,time,external) VALUES (?,?,?,?,?,?,?,?)" );
+
```
bobzilladev: Why make a different statement at all? Can just use the regular insertLocation but put external on the end, and make sure all callers set it properly (there pry is only 1 or 2?).

rksh: least-risk - `external` defaults to zero - we make the smallest change to the code path by using the old statement and relying upon the column default.

rksh: updated

PR #404, opened 2020-02-16
[ENH] simplifying endless scroll into a new PR

rksh:
The old PR was a mess - simplifying for inclusion.

### DIFF
--- a/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/UploadsFragment.java
+++ b/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/UploadsFragment.java
@@ -195,6 +195,7 @@ public class UploadsFragment extends Fragment {
         if (!busy) {
             final int pageStart = page * ROW_COUNT;
             final String downloadUrl = MainActivity.UPLOADS_STATS_URL + "?pagestart=" + pageStart + "&pageend=" + (pageStart + ROW_COUNT);
+            busy = true;
             final ApiDownloader task = new ApiDownloader(getActivity(), ListFragment.lameStatic.dbHelper, null,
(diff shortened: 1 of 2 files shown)


wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/UploadsFragment.java
```diff
@@ -176,29 +182,39 @@ private void setupSwipeRefresh(final LinearLayout rootView) {
...
-                    @Override
-                    public void requestComplete(final JSONObject json, final boolean isCache) {
-                        handleUploads(json, handler);
-                    }
-                });
-        try {
-            task.startDownload(this);
-        } catch (WiGLEAuthException waex) {
-            MainActivity.info("Transactions Download Failed due to failed auth");
+        if (!busy) {
```
bobzilladev: not seeing where busy is set to true?

rksh: oops.

bobzilladev: should probaby be an atomic boolean test and set, but humans are slow :)

rksh: I was a little concerned about that, since AtomicBooleans appear to persist in weird ways on Android. I could force-set it on instantiation, would probably make it safe.

rksh: should we reset the atomicBoolean on-create or on-resume to prevent locking ourselves out of performing downloads?

PR #420, opened 2020-06-04
altering WPA3 test to include SAE, OWE, EAP_SUITE_B_192

rksh:
these don't really address RSN-EAP-CCMP yet, but this will be closer.
new paths not exercised yet due to lack of wpa3 nets in test areas.

### DIFF
--- a/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/model/Network.java
+++ b/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/model/Network.java
@@ -33,6 +33,10 @@ public final class Network implements ClusterItem {
     private static final String BAR_STRING = " | ";
     private static final String DASH_STRING = " - ";
     private static final String WPA3_CAP = "[WPA3";
+    private static final String SAE_CAP = "SAE";
+    private static final String SUITE_B_192_CAP = "EAP_SUITE_B_192";
+    private static final String OWE_CAP = "OWE";    //ALIBI: handles both OWE and OWE_TRANSITION
+    //TODO: how we do distinguish between RSN-EAP-CCMP WPA2 and WPA3 implementations?
     private static final String WPA2_CAP = "[WPA2";
     private static final String RSN_CAP = "[RSN";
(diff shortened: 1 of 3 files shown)


wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/model/Network.java
```diff
@@ -33,6 +33,10 @@
     private static final String BAR_STRING = " | ";
     private static final String DASH_STRING = " - ";
     private static final String WPA3_CAP = "[WPA3";
+    private static final String SAE_CAP = "SAE";
+    private static final String SUITE_B_192_CAP = "EAP_SUITE_B_192";
+    private static final String OWE_CAP = "OWE";    //ALIBI: handles both OWE and OWE_TRANSITION
```
bobzilladev, line 38: The SAE and OWE don't have any anchor punctuation, which makes it more likely we'll get a bad conflict in the future. Can we add a hyphen or full prefix like "[RSN-SAE"? Would rather lock down now than be sad by clients in the field later.

rksh: This was because the actual set of strings is quite large/varied; I was using the google and other implementations as references.
Not sure how to find a complete list; documentation is inconsistent

bobzilladev: could we at least have '-SAE' and '-OWE'? Or is it truly unbounded?

rksh: I _think_ we could add `-` in front of each.
Android itself is equally agnostic:
https://android.googlesource.com/platform/frameworks/base/+/master/packages/SettingsLib/src/com/android/settingslib/wifi/AccessPoint.java#1708

bobzilladev: loller cats, well, who am i to argue.

PR #422, opened 2020-06-09
[FIX] fixing up hash keys for m8b export

rksh:
converting sipkey hashables to long (avoiding negative ints inherent in signed int32)
Doing this right would involve a test suite.


bobzilladev:
loooooooong

PR #432, opened 2020-09-12
[ENH] superimpose distance on map

rksh:
just add ft/mi or m/km with distance on the map.
also fixes a noise-based distance creep issue in distance recording


wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/listener/GPSListener.java
```diff
@@ -38,6 +38,9 @@
     public static final long MIN_ROUTE_LOCATION_DIFF_TIME = 3 * 1000;
     public static final float MIN_ROUTE_LOCATION_PRECISION_METERS = 24.99f;
 
+    //Minimum difference between updates to change total distance.
+    public static final float DIST_MIN_T = 0.5f;
```
bobzilladev, line 42: This is a factor of refresh too, so if someone is updating gps twice a second this would drop everything while walking under a meter per second. If this is to reduce the amount of updates on the map side, maybe put the throttle over there, instead of introducing error here?

bobzilladev (approved):
I love it. MAX_ENGINEERING

PR #436, opened 2020-09-17
[ENH] 

rksh:
modernizing the network map display:
 - move filter tools to the bottom so they're more reachable on hand-help devices
 - make the data bar a translucent overlay
 - fix blank screens by trilaterating a location in software


bobzilladev:
the red ring is the current 'network' table view of the network, so the latest highest-signal observation. doesn't use all observations like your trilateration, so that will probably maybe do a better jorb.

bobzilladev (approved):
Seems okally dokally, can get it merged and play with.

PR #437, opened 2020-09-18
[ENH] making the map observation markers zoom-proof

rksh:
using geo-shape circles in the map network detail created circumstances where zooming in and out would render the observations too small to see or large enough to interfere with the map. (https://github.com/wiglenet/wigle-wifi-wardriving/pull/436 was highlighting this by dynamically zooming in/out) Using markers derived from a vector asset remedies this problem, as well as giving us "click targets" if we want to provide detail of individual observation properties in future releases.


bobzilladev:
how well does that scale? we show up to like 1024 observations don't we?

bobzilladev (approved):
out of band: blit fasterer than drawin, gdg

Issue #438, opened 2020-09-20
Bug: tts speaks in english instead of german


Dtrieb:
My Device and tts language is german, but wigle speaks English, tts is english.

I want to set it p.e. to german.

rksh:
interesting - this should work. can you share device, OS version with us? Specifically, some FOSS, 3rd party Android versions probably won't have the support we need.

Dtrieb:
Phone: Google Pixel 4 XL
OS: Stock Android 11
Nothing else special ...

The language in the Wigle is set to german.

rksh:
working on a solution in https://github.com/wiglenet/wigle-wifi-wardriving/pull/440 - watch this space.

rksh:
tested - resolved in https://github.com/wiglenet/wigle-wifi-wardriving/pull/440 - marked for release 2.56

rksh:
released in new betas! closing.

PR #440, opened 2020-09-23
[ENH] here's a way to download speech packs and switch...

rksh:
should address https://github.com/wiglenet/wigle-wifi-wardriving/issues/438

note: this can't send the update intent from the static method if the context is null - would appreciate feedback on how i'm overcomplicating this.


wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/MainActivity.java
```diff
@@ -2674,4 +2714,24 @@ public void endRouteMapping(SharedPreferences prefs) {
...
+                state.tts.setLanguage(locale);
+            } else {
+                error("preferred locale: [" +locale+"] not available on device.");
+            }
+        } else if (status == TextToSpeech.SUCCESS) {
+            error("TTS init successful, but TTS engine was null");
+        } else if (status == TextToSpeech.ERROR) {
+            error("TTS init failed: "+status);
+        }
+    }
```
bobzilladev, line 2736: do we need a super.onInit? That's a common pattern, so making sure.

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/MainActivity.java
```diff
@@ -1665,6 +1694,10 @@ private void setupSound() {
                 info("cannot get call state, will play audio over any telephone calls: " + ex);
             }
         }
+
+        Intent checkTTSIntent = new Intent();
+        checkTTSIntent.setAction(TextToSpeech.Engine.ACTION_CHECK_TTS_DATA);
+        startActivityForResult(checkTTSIntent, ACTION_TTS_CODE);
```
bobzilladev, line 1700: this block shows up twice, might DRY it

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/MainActivity.java
rksh: that's a good question; documentation is pretty sparse - e.g. https://developer.android.com/reference/android/speech/tts/TextToSpeech.OnInitListener . it *looks* like it's an external signal on completion of TTS initialization (and this is borne out in testing). Searching sample code online doesn't turn up any super() examples.

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/MainActivity.java
rksh: i'd left these independent because the static version has to check to see whether the MainActivity has been instantiated yet. Perhaps I can refactor this.

bobzilladev (approved):

PR #442, opened 2020-09-26
[ENH] Xlation placeholders

rksh:
just long-outstanding missing lines in these translations.
google translate versions, probably terrible, but answering the push for more translation.


bobzilladev (approved):
This must have taken forever. I still vote we just move to exclusively Klingon.

PR #443, opened 2020-09-28
[ENH] Additional lang

rksh:
Language files contain all rows for every currently supported language. Translations may not be complete, but lines are present.


bobzilladev (approved):
..........klingon only

PR #444, opened 2020-09-29
[ENH] Cleanups from play release warnings

rksh:
the play console flags all kinds of crap. most doesn't matter.


bobzilladev (approved):
must obey our static analysis masters

PR #445, opened 2020-10-01
[FIX] it's always italian

rksh:


wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/background/AsyncDetailHull.java
```diff
@@ -0,0 +1,4 @@
+package net.wigle.wigleandroid.background;
+
+public class AsyncDetailHull {
+}
```
bobzilladev, line 4: an empty class?

rksh: oh dammit, auto-add is insidious

rksh:
re-issuing without garbage.

PR #451, opened 2020-10-31
Update Hungarian translation

meskobalazs:
I have translated the missing localized strings, and fixed some problems with the existing translation.

On acceptance, tagging as hacktoberfest-accepted would be most appreciated.


bobzilladev:
Wow, this was a ton of work, thank you!

PR #453, opened 2020-11-15
[ENH] fix for missing TTS intent activity on LG and BBerry Key2

rksh:
bug report courtesy of 'dd'
only a problem for specific handset ROMs:

https://stackoverflow.com/questions/25830230/android-text-to-speech-throws-activitynotfoundexception


bobzilladev (approved):

uhtu (commented):
👍

PR #454, opened 2020-11-15
[FIX] clear one non-testable ROM-specific bug, find the next non-testable ROM-specific bug.

rksh:
Interactive testing in the beta channel. :/
Thanks to dd for helping diagnose these.


bobzilladev:
try {
android();
} catch (Throwable meh) {
}

bobzilladev (approved):

Issue #379, opened 2019-10-24
Android Auto initilization silently closes/prempts wigle app process.


cognitive-rift:
Android 10 with Latest Android Auto and Wigle application on Pixel 3.

Step 1. Open Wigle wardriving app, let it fade into background.

Step 2. Turn on vehicle that Android Auto is paired with.

Step 3. As soon as Android Auto goes into driving mode, the Wigle app process is closed.

Expected behavior is that the Wigle app will run continuously until closed. In reality, wigle has to be re-opened every time driving mode is activated.

bobzilladev:
We have never tested with android auto, so it's a miracle it works at all! I'll link this to your other issue and close this one.
https://github.com/wiglenet/wigle-wifi-wardriving/issues/272

KoalaBear84:
This one should be opened again. It's not running ONTO Android Auto, it is crashing when Android Auto starts. Short after the display is taken over by the car display. And because of this bug I missed a lot of Wifi, because I thought it was started, but it had silently crashed.

Where can I find the (crash) logs (if any)? So I can update #272 

Thanks!

rksh:
I think we have a candidate for this - seeing onDestroy getting called during theme switch even without Android Auto - very likely the same issue.

KoalaBear84:
That is some great news. 👍

KoalaBear84:
Any idea when this will be released?

rksh:
This is now released - testing and feedback welcome

KoalaBear84:
Tested it today, it works! Wigle stays alive 👍

PR #457, opened 2020-12-21
[FIX] hopefully fixing #379

rksh:
Tested in android 10, 11. Seems to prevent crash on connect/disconnect from car.


bobzilladev (approved):
ok!

PR #458, opened 2020-12-24
[FIX] jb2 and kitkat don't like the observation markers

rksh:
apparently synthesizing these from vectors in JB2/KitKat hates the new standard.


rksh:
should fix the current low-version-number crashes.

bobzilladev (approved):
It's More Fun to Compat
Doot Doot Doo Doot

Issue #460, opened 2020-12-24
A display issue


franklinbill:
Dear developer!
In the process of using, I found a bug in layout / display. As shown in the figure, in the first page, the text of is blocked, which may be an adaptation problem. It is estimated that the height of the text component needs to be set to be adaptive to the text size.
App version is 2.51, my mobile phone is Google nexus 6, screen resolution setting is 1920* 1080, _**font size and display size**_ are the **_largest_** ! ! !, Android system is 8.0.
The above is a description of the problem. The overall experience of the software is good. You can see that the design is very attentive. Thank you very much for the efforts of the developers all the time. I hope to receive a reply.
Thank you!
![bug-wigleandroid-2](https://user-images.githubusercontent.com/18572115/103098572-31acdf00-4646-11eb-8950-b2eeb4de4da3.png)

rksh:
weird; we're not even doing android responsive layout wrong in that view. ;) Will test and see what's up.

rksh:
oh, here we are. https://stackoverflow.com/questions/11243867/android-last-line-of-textview-cut-off - don't have a Nexus 6 with which to test, but the behavior seems ok on other test devices with this fix applied.

rksh:
fix now released! thanks for the report!

Issue #459, opened 2020-12-24
A display bug


franklinbill:
Dear developer!
In the process of using, I found a bug in layout / display. As shown in the figure, in the first page, the lower part of the text overlaps with each other, which may be an adaptation error. It is possible that the text and display are too large.
App version is 2.51, my mobile phone is Google nexus 6, screen resolution setting is 1920* 1080, _**font size and display size**_ are the **_largest_** ! ! !, Android system is 8.0.
The above is a description of the problem. The overall experience of the software is good. You can see that the design is very attentive. Thank you very much for the efforts of the developers all the time. I hope to receive a reply.
Thank you!
![wigleandroid](https://user-images.githubusercontent.com/18572115/103098452-c2cf8600-4645-11eb-8c3d-1da6a3e368b9.png)

rksh:
fix now released - thanks for the report!

PR #467, opened 2021-01-22
[FIX] accessibility fixes for #460 and #459, fixing an incorrect log message

rksh:
Fixing the large-font accessibility issues reported in #460 and #459 .
Fixing an incorrect log message.


bobzilladev (approved):
bye bye baseline align

PR #468, opened 2021-01-24
[FIX] day/night on SDK28+

rksh:
this is an attempt to get a clean UI restart on theme change without breaking everything. some of this is just de-crufting.

- TTS setup was launching on each UI re-instantiation because it was part of setup sound; does the setLocale call above it in onCreate handle this?
- setLocal was being called twice in onCreate (?!)
- adding an atomicBoolean to prevent service termination on a simple UI start; feels like the elaboration of a bad idea, but the alternative is to completely unwire our service management stuff from MainActivity which seems equally fraught
- some linting just because MainActivity is a file that's been through tons of changes over the course of a decade, wanted to see what the real warnings were.


bobzilladev (approved):
seems ok. lifecycle management in android is such spaghetti.

PR #469, opened 2021-01-24
[ENH] TTS + configuration change

rksh:
ok, this seems to fix TTS activation AND configuration change.


bobzilladev (approved):
nothing horrific, taking your word about fixing :)

PR #477, opened 2021-03-09
[FIX] Bad token prevention

rksh:
Toast and Dialog presentation apparently causes occasional BadTokenExceptions - this should prevent app termination out from under display of modals.

### DIFF
--- a/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/MappingFragment.java
+++ b/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/MappingFragment.java
@@ -34,6 +34,7 @@ import androidx.core.app.ActivityCompat;
 import androidx.fragment.app.DialogFragment;
 import androidx.fragment.app.Fragment;
 import androidx.core.view.MenuItemCompat;
+import androidx.fragment.app.FragmentActivity;
 
 import android.util.Base64;
 import android.view.LayoutInflater;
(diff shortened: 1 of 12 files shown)


bobzilladev (approved):
You would think they would have this check internally. Could we move the check to a helper method instead, passing in the interface that has isFinishing? Might be a little cleaner, remove the 'if' everywhere. Otherwise +1.

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/MappingFragment.java
```diff
@@ -362,13 +363,10 @@ public Tile getTile(int x, int y, int zoom) {
 
                             try {
                                 final byte[] data = downloadData(new URL(s), userAgent, authToken);
-                                if (data != null) {
-                                    return new Tile(providerTileRes, providerTileRes, data);
-                                }
+                                return new Tile(providerTileRes, providerTileRes, data);
```
bobzilladev, line 366: data can actually never be null?

rksh: according to the linter!

bobzilladev: obey

bobzilladev (approved):

PR #480, opened 2021-03-11
[FIX] working on fixing concurrentModification issue from play console

rksh:
As reported:

```
java.util.ConcurrentModificationException: 
  at java.util.ArrayList$Itr.next (ArrayList.java:860)
  at net.wigle.wigleandroid.UploadsFragment$RankDownloadHandler.handleMessage (UploadsFragment.java:282)
  at android.os.Handler.dispatchMessage (Handler.java:107)
  ```

### DIFF
--- a/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/UploadsFragment.java
+++ b/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/UploadsFragment.java
@@ -105,8 +105,7 @@ public class UploadsFragment extends Fragment {
     private static final int ROW_COUNT = 100;
 
     private int currentPage = 0;
-    private final ArrayList<Parcelable> resultList = new ArrayList<>(ROW_COUNT);
-    private AtomicBoolean busy = new AtomicBoolean(false);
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by uhtu — APPROVED** [review]
👍

File: wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/UploadsFragment.java [hunk 1]
```diff
@@ -290,10 +291,10 @@ public void handleMessage(final Message msg) {
                         uploadsListAdapter.add(upload);
                     }
                 }
-                final SwipeRefreshLayout swipeRefreshLayout =
-                        view.findViewById(R.id.uploads_swipe_container);
-                swipeRefreshLayout.setRefreshing(false);
             }
+            final SwipeRefreshLayout swipeRefreshLayout =
+                    view.findViewById(R.id.uploads_swipe_container);
+            swipeRefreshLayout.setRefreshing(false);
```
bobzilladev, line 297: Feels like this should at least be in a check that the msg is MSG_RANKING_DONE. If we add more message types it may not make sense to set refreshing off here. But we could always leave that on whoever adds another msg to deal with.

PR #481, opened 2021-03-11
[ENH] interrupt TTS on any TTS option disable

rksh:
remedy for #81 - when TTS checkboxes are "unticked," in-progress TTS is interrupted.
If SSID is the only problem, we *could* limit this to just that checkbox, but this is a simple fix.


bobzilladev (approved):
SHUT UP

PR #487, opened 2021-06-15
removing OWE/OWE_TRANSITIONAL from the WPA3 cases.

rksh:
Thanks to @ZeroChaos- - removing OWE/Transition from WPA3 tests on client display.


bobzilladev:
Seems like we should add an OWE bucket server side, then add OWE bucket client side, rather than just remove.

ZeroChaos-:
for now just removing owe from wpa3 makes sense.  I'll try to capture some data for @rksh to better support OWE and wpa3

bobzilladev:
Nevermind, reading internal chat :)

rksh:
nothing's being removed - the full capabilities list is still getting sent to the server in AuthType. I'm just striking OWE from the list of things that result in "WPA3" categorization on the client. Discussion suggests that we might consider holding a Set of OWE pointers on the client and adding a visual indicator of the pairing.

bobzilladev (approved):

PR #490, opened 2021-07-08
[FIX] trying to fix a race condition.

rksh:
trying to lock down causes of occasional ConcurrentModification problems with the uploads list.


uhtu (approved):

bobzilladev (approved):
my lovely lady locks

Issue #475, opened 2021-02-20
dark mode bug


clientenq:
Since the latest version of wigle on android the list frequently enters dark mode when darkmode is off.
This is only related to scan results text. The light/white text color changes to dark/dark gray.
Result is barely readable entries, since it is on a dark background.
List icons and everything above the list ( location, altitude, totals etc) are not affected.

I call this swiching to dark mode since it is related to device dark mode. That is turning device dark mode on and off seems to resolve the issue when it occurs.

Expected behaviour: colour of the mendioned text should only change to dark when requested.

Bug occurs since latest version. Wigle 2.60. No other entries or features seem to be affected.

rksh:
Interesting - can you share OS version and device type?
2.60 introduced theme support but left it deactivated in testing - clearly there's a combination where the system is overriding overriding our hardcoded theme.
1. does exiting and restarting the app make a difference?
2. does this happen during runs when dark/light mode change (i.e. while using Android Auto?)

Thanks for the report, sorry for the difficulty.

rksh:
2.61 will now add a preferences toggle to allow users to select light/dark/follow system, however we'd love to figure this out beforehand!

rksh:
I can reproduce this on Android 10 / a Samsung A10 - looking into remediation.

rksh:
it appears to me that the issue can be reproduced on the A10 by changing the device orientation to landscape, then back to portrait. this flows through 
`onCreate` and `onConfigurationChanged` in `MainActivity` with the correct display modes set, but the re-instantiated UI still shows the "day" text color in the list view.

rksh:
believe this is fixed in the upcoming release. If not, please include screenshots and device information!

PR #491, opened 2021-07-10
[FIX] List view fix

rksh:
some samsung phones aren't respecting light text in list view in night mode. Remedies https://github.com/wiglenet/wigle-wifi-wardriving/issues/475

### DIFF
--- a/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/NetworkActivity.java
+++ b/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/NetworkActivity.java
@@ -65,6 +65,7 @@ import net.wigle.wigleandroid.model.Network;
 import net.wigle.wigleandroid.model.NetworkType;
 import net.wigle.wigleandroid.model.OUI;
 import net.wigle.wigleandroid.ui.NetworkListUtil;
+import net.wigle.wigleandroid.ui.ThemeUtil;
 
 @SuppressWarnings("deprecation")
 public class NetworkActivity extends AppCompatActivity implements DialogListener {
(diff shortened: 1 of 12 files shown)


wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/NetworkActivity.java
```diff
@@ -102,6 +106,17 @@ public void onCreate(Bundle savedInstanceState) {
...
+        final int displayMode = prefs.getInt(ListFragment.PREF_DAYNIGHT_MODE, AppCompatDelegate.MODE_NIGHT_YES);
+        final int nightModeFlags = getResources().getConfiguration().uiMode & Configuration.UI_MODE_NIGHT_MASK;
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
+            if (AppCompatDelegate.MODE_NIGHT_YES == displayMode ||
+                (AppCompatDelegate.MODE_NIGHT_FOLLOW_SYSTEM == displayMode &&
+                        nightModeFlags == Configuration.UI_MODE_NIGHT_YES)) {
+                getWindow().setNavigationBarColor(WindowManager.LayoutParams.FLAG_TRANSLUCENT_NAVIGATION);
+            }
+        }
+
```
bobzilladev, line 119: Can this call setNavTheme too?

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/NetworkActivity.java
```diff
@@ -436,7 +451,7 @@ public static float cleanSignal(Float signal) {
         return signalMemo;
     }
 
-    private void setupMap( final Network network, final Bundle savedInstanceState ) {
+    private void setupMap( final Network network, final Bundle savedInstanceState, final SharedPreferences prefs ) {
```
bobzilladev, line 454: Doesn't look like this new prefs arg is used.

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/NetworkActivity.java
rksh: experimentation suggests that calling into MainActivity doesn't affect the NetworkActivity's navbar (and since it needs to get context, can't be done as a static call)

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/NetworkActivity.java
rksh: ah, meant to leave a comment here; this is pre-plumbing to extend day/night mode control to the maps as a future feature.

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/NetworkActivity.java
bobzilladev: Could pass in the context like we do all over the place to avoid the duplicated code?

bobzilladev (approved):
Weird that you have to grab the window first, but it's Android so of course, thanks for DRYin' like a boss.

PR #492, opened 2021-07-15
[FIX] proposed fix for play console exception - needs review

rksh:
there's a narrow window in the run loop for DatabaseHelper where a request gets issued, but blows up because we've terminated the app - we can choose not to freak out if the error happened because of asynchrony during shutdown, but I'm not positive that this is a wholly harmless change, would appreciate insight into the original design.


bobzilladev (changes requested):
That seems like the right thing to do. If we're exiting so be it. i would at least log a warn though so it's not silently eaten.

rksh:
excellent point, added.

bobzilladev (approved):

PR #493, opened 2021-08-16
[FIX] a stab at improving our channel stuff.

rksh:
a problem detected by @Czechball with our outdated and maybe buggy channel mapping


wiglewifiwardriving/src/test/java/net/wigle/wigleandroid/model/NetworkChannelTest.java
```diff
@@ -0,0 +1,268 @@
...
+            //I can find no documentation regarding channels 237-255
+            freqToChanTemp.put(2312 + 5 * (i - 237), i);
+        }
+
+        for (int i = 0; i <= 13; i++) {
+            freqToChanTemp.put(2407 + (5 * i), i);
+        }
+        freqToChanTemp.put(2484, 14);
+
+        freqToChanTemp.put(5170, 34);
```
ZeroChaos-, line 75: why hardcode all these and not use the code from linux to (mostly) do the math?

rksh: per the comments, this is the code moved OUT of Network.java into the test suite in order to compare new/old. functionality. It is replaced by the linux-like functionality in the static methods on Network.java in this changeset.

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/model/Network.java
```diff
@@ -301,12 +255,79 @@ public boolean equals(final Object other) {
...
             }
         }
-        return null;
+        switch (band) {
+            case WIFI_24_GHZ:
+                if (channel == 14) {
+                    return 2484;
+                } else if (channel < 14) {
+                    return (2407 + channel * 5);
+                }
```
jmberg, line 280: I think you need "break;" in all these cases when you didn't find anything correctly, or "return null;" maybe, otherwise you'd fall through to the next case which would be ... really weird?

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/model/Network.java
```diff
@@ -301,12 +255,79 @@ public boolean equals(final Object other) {
...
+                if (channel <= 233) {
+                    return 5950 + channel * 5;
+                }
+            case WIFI_60_GHZ:
+                if (channel < 7) {
+                    return 56160 + channel * 2160;
+                }
+            case WIFI_S1GHZ:
+                return 902000 + channel * 500;
```
jmberg, line 300: you're getting the units confused here - S1GHz is "sub 1 GHz" and that's in kHz not MHz, but all the other cases you're using MHz, not KHz...

In Linux all of this is in MHz, but we use MHZ_TO_KHZ() in all but the S1G case.

maybe you don't want to handle S1G at all if you're doing things in MHz still, or otherwise you have to covert everything to KHz, or start using floats (1/2 can be represented exactly), though not sure I'd recommend that

jmberg: so really - in MHz - this should be "902 + channel * 1/2" but you can't express that in integers

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/model/Network.java
rksh: good point - adding.

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/model/Network.java
```diff
@@ -301,12 +255,79 @@ public boolean equals(final Object other) {
...
+        if (frequencyMHz == 2484) {
+            return 14;
+        } else if (frequencyMHz < 2484) {
+            return (frequencyMHz - 2407) / 5;
+        } else if (frequencyMHz >= 4910 && frequencyMHz <= 4980) {
+            return (frequencyMHz - 4000) / 5;
+        } else if (frequencyMHz < 5925) {
+            return (frequencyMHz - 5000) / 5;
+        } else if (frequencyMHz == 5935) {
+            return 2; //WTF, linux. looks like 5GHz / 187 to me.
```
jmberg, line 326: nothing to do with Linux, this is the spec - and that channels is considered on the 6 GHz band, not 5 GHz band.

Check Annex E.

wiglewifiwardriving/src/test/java/net/wigle/wigleandroid/model/NetworkChannelTest.java
```diff
@@ -0,0 +1,268 @@
...
+    //static final List<Integer> wiFi60GHzChannelNumbers = Arrays.asList(new Integer[] {1, 2, 3, 4, 5, 6, 9, 10, 11, 12, 13, 17, 18, 19, 20, 25, 26, 27});
+
+    //ALIBI: this will be both unit tests for the new code, and edge case detection for the old.
+    //START: bad old code
+    private static final Map<Integer, Integer> freqToChan;
+    static {
+        Map<Integer, Integer> freqToChanTemp = new HashMap<>();
+        for (int i = 237; i <= 255; i++) {
+            //I can find no documentation regarding channels 237-255
+            freqToChanTemp.put(2312 + 5 * (i - 237), i);
```
bobzilladev, line 67: This is an encoding for cell frequencies that we do see. In the new code 2312 will come out as channel -19, but we store as unsigned, so maybe we keep this crazy scheme going forward to keep this information rather than it being clamped to zero. The right answer is probably storing the raw frequencies which we get from clients now a days but didn't in 2001, plumbing that through everything will be quite a chore.

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/model/Network.java
rksh: great catch - sloppy porting on my part.

wiglewifiwardriving/src/test/java/net/wigle/wigleandroid/model/NetworkChannelTest.java
```diff
@@ -0,0 +1,268 @@
...
+            5825, 5835, 5845, 5855, 5865, 5875, 5885, 5900/*, 5910, 5915,
+            5920, 5935, 5940, 5945, 5960, 5980*/);
+
+    static final List<Integer> wiFi5GHzChannelNumbers = Arrays.asList(7, 8, 9, 11, 12,
+            16, 32, 34, 36, 38, 40, 42, 44, 46, 48,
+            50, 52, 54, 56, 58, 60, 62, 64, 68, 96,
+            100, 102, 104, 106, 108, 110, 112, 114, 116, 118,
+            120, 122, 124, 126, 128, 132, 134, 136, 138, 140,
+            142, 144, 149, 151, 153, 155, 157, 159, 161, 163,
+            165, 167, 169, 171, 173, 175, 177, 180/*, 182, 183,
```
jmberg, line 41: 180 really shouldn't be valid according to Annex E, have you seen that in the wild somehow?

wiglewifiwardriving/src/test/java/net/wigle/wigleandroid/model/NetworkChannelTest.java
jmberg: there aren't any channels 237-255, 6 GHz only goes up to 233, 2.4 GHz only has 1-14, and 4 GHz only has 36-177 (or sometimes 179?)

wiglewifiwardriving/src/test/java/net/wigle/wigleandroid/model/NetworkChannelTest.java
rksh: no, the cutoff in the test cases was just predicated on notes to the effect that channels 182-and-up were unimplemented - 180 was an ambiguous case.

rksh: no reports of 180 in the wild - however looking at https://wigle.net/stats#octetstats , we do have reports of small numbers of channels 183, 184, 185(?!), 187, 188, 189, 196(?!).

jmberg: IIRC, even in Linux some early versions of 6 GHz (WiFi 6E) support might have used some of those channel numbers, I guess you don't have stats on the actual frequencies?

187 is 5935 MHz, which is 6E channel 2.
The others don't really map well to 6E, though 6E channelization also changed a few times, and I don't remember all the details now.

rksh: we should have designed out model around frequencies instead of channels 20 years ago, hard to find actual frequencies. I've also found some old specs indicating mappings for the weird channel numbers down into 4.9GHz. Since we have so much old data, we have a conundrum.

wiglewifiwardriving/src/test/java/net/wigle/wigleandroid/model/NetworkChannelTest.java
rksh: WiFiBand -> NetworkBand and added a Cell 23 GHz enum entry. not ideal, but it will maintain the old functionality.

jmberg: There "23 GHz" seems odd though, did you mean "4 GHz" or something? Whatever it means.

Note that there *are* actually entries in Annex E table E-4 (Global operating classes) for such a thing:

image
image

so maybe some of those were in use - but I don't know of many devices supporting them.

jmberg: and also, 5/10 MHz channels aren't typically used in WiFi APs, other than special-purpose installations, and not supported by clients either, so you can probably ignore those entries

rksh: the frequency values with that weird old hack would map [237, 238, 239, 240, 241, 242, 243, 244, 245, 246, 247, 248, 249, 250, 251, 252, 253, 254, 255] to [2312, 2317, 2322, 2327, 2332, 2337, 2342, 2347, 2352, 2357, 2362, 2367, 2372, 2377, 2382, 2387, 2392, 2397, 2402] - per @bobzilladev this mapping actually gets used for cells.

bobzilladev (approved):
ok by me

PR #499, opened 2021-09-15
Fix background scanning

CoD-Segfault:
Adding background location permissions seems to resolve the issue where WiFi scanning does not occur when the app is not in focus.


CoD-Segfault:
Just want to add that it affects more than just the Pixel 3A, as it's happening on my OnePlus 6 running LineageOS 18.1.  The forum post shows all of my findings leading up to this workaround.

bobzilladev:
Fix for the root cause looks to be: https://github.com/wiglenet/wigle-wifi-wardriving/pull/500

PR #500, opened 2021-09-16
As of september 2021 patch android now considers calling startForegro…

bobzilladev:
…und on a service that is already foreground as starting from background, add a guard to call once.


uhtu (commented):
👍

PR #502, opened 2021-12-04
[ENH] Gpx browser

rksh:
Want access to logged routes? Here you go!
Note: option only appears if route logging is on.
Not applicable for FOSS release.

### DIFF
--- a/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/DataFragment.java
+++ b/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/DataFragment.java
@@ -13,6 +13,7 @@ import net.wigle.wigleandroid.db.DBException;
 import net.wigle.wigleandroid.db.DatabaseHelper;
 import net.wigle.wigleandroid.model.Pair;
 import net.wigle.wigleandroid.ui.WiGLEToast;
+import net.wigle.wigleandroid.util.AsyncGpxExportTask;
 import net.wigle.wigleandroid.util.FileUtility;
 import net.wigle.wigleandroid.util.MagicEightUtil;
 import net.wigle.wigleandroid.util.SearchUtil;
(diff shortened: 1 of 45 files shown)


bobzilladev (approved):
Wow, you've been busy. All looks cogent to me, except the recyclerview stuff which I assume is Dark Alchemy that just must be done.

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/DataFragment.java
```diff
@@ -997,166 +995,4 @@ protected void onProgressUpdate(Integer... values) {
             }
         }
     }
-    class AsyncGpxExportTask extends AsyncTask<Long, Integer, String> {
```
uhtu, line 1000: mmm refactorsticking

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/util/AsyncGpxExportTask.java
```diff
@@ -0,0 +1,215 @@
...
+            //}
+            final double lat = cursor.getDouble(0);
+            final double lon = cursor.getDouble(1);
+            final long time = cursor.getLong(2);
+
+            writer.append("<trkpt lat=\"").append(String.valueOf(lat)).append("\" lon=\"")
+                    .append(String.valueOf(lon)).append("\"><time>").append(
+                    dateFormat.format(new Date(time))).append("</time></trkpt>\n");
+            lineCount++;
+            if (totalCount == 0) {
```
uhtu, line 142: i know this is just a refactor move, but that seems like a strange case, ne?

uhtu (approved):

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/DataFragment.java
rksh: now to move all the other asyncTasks out

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/util/AsyncGpxExportTask.java
rksh: probably belt and suspenders. this one will need a hard look to make sure it works right everywhere.

PR #509, opened 2022-02-03
[FIX] fixing a deps problem and remove-extension fxn

rksh:
library version bump and behavior fix from conversation with uhtu


bobzilladev (approved):
-1 is not the loneliest number

PR #514, opened 2022-03-12
[ENH] Determining the coverage area of a given access point

Krzysztofz01:
The functionality that I would like to implement is the ability to save information in which location a given network has the lowest signal level and to overwrite the default location, if the signal turned out to be stronger. Having these two pieces of information will allow to estimate the area in which a given network is available. One of my projects is able to properly parse data from the wigle scanner and adding such functionality would be very useful both for the wigle platform itself and for other independent projects.

My branch currently does not have the support for network CSV export implemented, also the new functionality has no unit test coverage yet.

### DIFF
--- a/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/listener/WifiReceiver.java
+++ b/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/listener/WifiReceiver.java
@@ -237,10 +237,27 @@ public class WifiReceiver extends BroadcastReceiver {
                 }
                 somethingAdded |= added;
 
-                if ( location != null && (added || network.getLatLng() == null) ) {
-                    // set the LatLng for mapping
-                    final LatLng LatLng = new LatLng( location.getLatitude(), location.getLongitude() );
-                    network.setLatLng( LatLng );
+                if (location != null) {
+                    final LatLng currentLocation = new LatLng(location.getLatitude(), location.getLongitude());
+                    
+                    // Set the current location if the current network location value is null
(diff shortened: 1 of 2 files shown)


rksh:
I completely dig adding adding more coverage tools, but a few notes on approach:

1.  I think the app may already have more powerful "coverage" data you can use.
2. This won't be persisted across multiple executions of the application or manage the irregular coverage that results from the complexity of signal propagation
3. This will be sensitive to a lot of driver and radio problems we filter out right now (that result in statistical outliers).

check out https://wigle.net/phpbb/viewtopic.php?f=5&t=2892&p=11057&hilit=propagation&sid=8ba61cc02f557d0b80de50a3ce869e08#p11057 for details on path propagation and the dangers of trying to assume things about signal strength and coverage

see https://github.com/wiglenet/wigle-wifi-wardriving/blob/master/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/NetworkActivity.java#L284 and specifically trace the `obsMap` variable for how to request all the viewed observations for a point.

The CSV output isn't meant to convey the coverage data (and we don't lightly accept changes to the CSV format, since it's a standard used by not just the WiGLE server as well,  but also many other parties https://api.wigle.net/csvFormat.html ) - instead, it's the "input" to the server side coverage and signal calculations. You're welcome to implement a parallel concave hull algorithm in the network view if you'd like - you can see these for points in the "Basic Search" tool as a logged-in user of the website here:

https://wigle.net/mapsearch

when you search for a network and select a network in the result list.
you can check out a sample coverage implementation by tracing the `silverlinings ` and `goldlinings` variables in

https://wigle.net/js/mapsearch.js

PR #516, opened 2022-03-21
[ENH] using new SDK30 is-throttled check when possible to improve message.

rksh:
we'll only show banner warnings in new releases now if dev mode isn't enabled / if throttle is on.


bobzilladev (approved):
Nice

PR #521, opened 2022-05-05
[FIX] moving logging calls out of MainActivity.

rksh:
Intent was to remove logging from over-burdened `MainAcitivty.java`

Since we end up needing `MainAcitivity` state to write to the logfile (and hence end up exposing `MainActivity.saveLog(String)` to call in each `Logging` method), is this really worth it? Does this just add risk by shuffling things around?


bobzilladev (approved):
1,016 line change to move 30 lines, that's the kind of massive effort for small gain that I'm here for! Good luck getting a change like this through in MegaCorp. Smell that? That's the smell of FREEDOM. My name is bobzilla, and I approve this commit.

PR #522, opened 2022-05-16
reversing a decade of forcing imperial units on the world.

rksh:
if you install in not-an-imperial-units country, default to metric.


bobzilladev (approved):
One thought is changing all the getters to use that function call as the default, but this covers all current and future cases, so I think it's ok dokey.

PR #523, opened 2022-05-20
[FIX] Route simplification threshold

rksh:
this is making dynamic mapping look wonky


bobzilladev (approved):
the only constants are change

ladaland (approved):

PR #524, opened 2022-06-07
[FIX[ Query prepared for network

rksh:
Switching from hand-concatenated to parameterized SQL in a quest to improve security and performance.

### DIFF
--- a/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/background/QueryThread.java
+++ b/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/background/QueryThread.java
@@ -22,9 +22,13 @@ public class QueryThread extends Thread {
     }
     public static class Request {
         private final String sql;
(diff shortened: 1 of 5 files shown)


ladaland (approved):

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/background/QueryThread.java
```diff
@@ -64,7 +66,7 @@ public void run() {
                 if ( request != null ) {
                     final SQLiteDatabase db = dbHelper.getDB();
                     if ( db != null ) {
-                        cursor = db.rawQuery( request.sql, null );
+                        cursor = db.rawQuery( request.sql, request.args );
```
bobzilladev, line 69: args can be null, will raw query handle that ok?

bobzilladev (approved):
lgtm

PR #527, opened 2022-07-15
[FIX] OUI availability check implementation

Krzysztofz01:
Minor change to the OUI class, which now includes an additional boolean value to indicate if *Properties* initialization was successful. Although *Properties* is initialized, it is partially protected against ["Partially initialized instances"](https://www.oracle.com/java/technologies/javase/seccodeguide.html#7-3). Moreover, I added a method that can check if OUI data is available. When *Properties* does not contain a given key, the return value is null, but keeping the information about data availability we do not have to rely on this mechanic.


bobzilladev:
The data returned will be null if the property isn't found, so this is logically equivalent to the existing code, so going to close this one.

PR #529, opened 2022-07-18
[ENH] Original SSID with commas valid CSV format fix

Krzysztofz01:
I do not know if this pull request will be approved because I assume that the compatibility with the WiGLE backend is key and it probably handles SSIDs in its own way (in which the commas have been replaced with underscores). Recently, while researching networks, I came across a large number of networks with SSIDs containing commas. The names looked something like this:
```
Hotspot 2,4Ghz
Hotspot 5,0Ghz
```
In one of my projects, I dealt with the appropriate SSID formatting on the backend, but here this problem can be solved during the object-CSV serialization stage. The solution is to put the SSID in quotation marks according to the [RFC4180](https://datatracker.ietf.org/doc/html/rfc4180#section-1):

_6.  Fields containing line breaks (CRLF), double quotes, and commas should be enclosed in double-quotes._

Thanks to this, the CSV format will be valid and we will maintain the highest level of compliance with the actual parameters of a given access point.

### DIFF
--- a/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/background/ObservationUploader.java
+++ b/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/background/ObservationUploader.java
@@ -479,8 +479,9 @@ public class ObservationUploader extends AbstractProgressApiRequest {
                 countStats.lineCount++;
                 String ssid = network.getSsid();
                 if (ssid.contains(COMMA)) {
-                    // comma isn't a legal ssid character, but just in case
-                    ssid = ssid.replaceAll( COMMA, "_" );
+                    // When the ssid is containing a comma, we must wrap the ssid with
+                    // quotes to ensure the output CSV format is valid
+                    ssid = String.format("\"%s\"", ssid);
                 }
                 // ListActivity.debug("writing network: " + ssid );
 


wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/background/ObservationUploader.java
```diff
@@ -479,8 +479,9 @@ private long writeFileWithCursor( final OutputStream fos, final Bundle bundle,
                 countStats.lineCount++;
                 String ssid = network.getSsid();
                 if (ssid.contains(COMMA)) {
-                    // comma isn't a legal ssid character, but just in case
-                    ssid = ssid.replaceAll( COMMA, "_" );
+                    // When the ssid is containing a comma, we must wrap the ssid with
+                    // quotes to ensure the output CSV format is valid
+                    ssid = String.format("\"%s\"", ssid);
```
rksh, line 484: This is correct, but rolling our own CSV might be the more "core" problem here! We're adding proper server side support for UTF-8 and CSV in the CSV format 1.5 upcoming release.... watch this space.

rksh (approved):

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/background/ObservationUploader.java
Krzysztofz01: I understand that the decision about implementing a custom CSV parser was due to avoiding third party dependencies? If external libraries are not a problem, I can implement a parser so that it can be used in the application in the future, when the backend is ready for it.

rksh: This PR is a great motivator for us - we've actually be doing a lot of work on the backend to prepare for a new CSV format version on our side - this coincides well with revisiting the parser.
We appreciate the PR, and "watch this space" - we'll merge once we're confirmed ready on the server-side!

XenoKovah:
Just as an FYI, I'm guessing it's already covered, but Bluetooth names can also have commas in them too

Krzysztofz01:
> [...] but Bluetooth names can also have commas in them too

As far as I know, if the `Network` in question is *Bluetooth/BT* then the device name is also stored in the `ssid` field. So this kind of CSV format escaping will also work for Bluetooth devices. Please correct me if I am wrong.

XenoKovah:
As far as I know too that's correct. But I figured a heads up was in order for the WiGLE folks to double check just in case it happened to get stored somewhere else too

Krzysztofz01:
Good point! Actually, it is not known how different types of networks are handled by the backend.

rksh:
I think this is getting incorporated/co-fixed in https://github.com/wiglenet/wigle-wifi-wardriving/pull/642
More importantly server-side parsing fixes are coming online so it'll work correctly.

Krzysztofz01:
I understand, in this case the patch in question will no longer be needed, I close the pull request.

Krzysztofz01:
> I think this is getting incorporated/co-fixed in #642 More importantly server-side parsing fixes are coming online so it'll work correctly.

Under such circumstances, do we still want to secure this on the mobile app side?

bobzilladev:
Aye, the mobile side is fixed in #642

PR #531, opened 2022-07-20
[ENH] Dashboard scan toggle support

Krzysztofz01:
The goal of this pull request is to implement the functionality proposed by users. Information about the scanning status from the dashboard will be replaced with a button that allows you to pause/resume the scanning process. I have refactored, also a function that formats the run and scan duration as suggested by _TODO_. I ran into a problem implementing `ConfirmationDialog`. I was guided by the code from `MainActivity`, but despite calling `ConfirmationDialog`, the `SharedPreference` does not change and such exception appears:
```
I/wigle: main] exception handling fragment alert dialog: java.lang.ClassCastException: net.wigle.wigleandroid.DashboardFragment cannot be cast to net.wigle.wigleandroid.DialogListener
    java.lang.ClassCastException: net.wigle.wigleandroid.DashboardFragment cannot be cast to net.wigle.wigleandroid.DialogListener
        at net.wigle.wigleandroid.MainActivity$ConfirmationDialog$1.onClick(MainActivity.java:942)
        at com.android.internal.app.AlertController$ButtonHandler.handleMessage(AlertController.java:172)
        at android.os.Handler.dispatchMessage(Handler.java:106)
        at android.os.Looper.loop(Looper.java:193)
        at android.app.ActivityThread.main(ActivityThread.java:6669)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:493)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:858)
```
I would be very grateful for suggestions on how to deal with this problem and get this pull request to be mergeable.

### DIFF
--- a/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/DashboardFragment.java
+++ b/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/DashboardFragment.java
@@ -37,6 +37,7 @@ import net.wigle.wigleandroid.util.Logging;
 import java.text.DecimalFormat;
 import java.text.NumberFormat;
 import java.util.Locale;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 public class DashboardFragment extends Fragment {
@@ -356,7 +357,7 @@ public class DashboardFragment extends Fragment {
                 PREF_QUICK_PAUSE,
                 QUICK_SCAN_PAUSE,
                 QUICK_SCAN_DO_NOTHING,
-                R.id.nav_list,
+                R.id.nav_dash,
                 QUICK_PAUSE_DIALOG);
     }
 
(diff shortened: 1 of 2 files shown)


rksh:
This one is a little charged - the dashboard screen is supposed to be display only and  proof against accidental interaction (there have also been requests for other controls such as upload.) This is why the user request wasn't implemented yet... thoughts @bobzilladev ?

bobzilladev:
Unfortunately people will accidentally press this and then have a horrible experience of not collecting anything on their drive, I think this needs to be a read-only view. Could add a series of confirmation dialog/checkboxes, but then folks will be annoyed accidentally pressing the button and having to clear a bunch of UI elements to get back to the driving view.

Krzysztofz01:
Right, sometimes you have to refuse to implement a desired features for the sake of the end-user. So I close the pull request, we can note in the *TODO* file that there is a pull request that was supposed to add such functionality.

PR #534, opened 2022-08-27
[ENH] 5g nr

rksh:
So, NR type Cells need to be supported.
Also generalizing some is-cell, is-bt, and is-GSM-like tests.

Don't merge until server-side support is complete.


wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/model/GsmOperator.java
```diff
@@ -153,9 +192,18 @@ public int getXfcn() {
...
+            case "LTE":
+            case "WCDMA":
+            case "GSM":
+                if ((cellId > 0) && (cellId < Integer.MAX_VALUE)) {
+                    return true;
+                }
+            case "NR":
+                if ((cellId > 0) && (cellId < Long.MAX_VALUE)) {
+                    return true;
+                }
```
bobzilladev, line 206: wonder if we can use the enum we have with all of these strings around, but not vital

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/model/NetworkType.java
```diff
@@ -33,4 +34,41 @@ public String getCode() {
...
+    public static final String channelCodeTypeForNetworkType(NetworkType type) {
+        switch (type) {
+            case GSM:
+                return "ARFCN"  ;
+            case LTE:
+                return "EARFCN";
+            case WCDMA:
+                return "UARFCN";
+            case NR:
+                return "NRARFCN";
```
bobzilladev, line 68: or bake these into the static enum generation as well

bobzilladev (approved):

PR #536, opened 2022-09-14
[FIX] fixing hints for things that aren't the maked-text input

rksh:
some weird old bugs have to do with these hints being at the wrong point in the layout.


bobzilladev (approved):
Odd that they want it at the layout rather than the input, but you do you android.

PR #539, opened 2022-10-05
Update German language strings

rumpelyann:
Update and fix German language strings


rksh:
AWESOME! Thank you for your help making the app better!

PR #545, opened 2022-11-12
[ENH] making reported run distance match route distance

rksh:
making the distance calc conditional match route addition - distances will match more closely.
TODO: eliminate low-accuracy jitter (still a problem)

### DIFF
--- a/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/ListFragment.java
+++ b/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/ListFragment.java
@@ -74,8 +74,6 @@ public final class ListFragment extends Fragment implements ApiListener, DialogL
     private static final int UPLOAD_DIALOG = 101;
     private static final int QUICK_PAUSE_DIALOG = 102;
 
-    public static final float MIN_DISTANCE_ACCURACY = 32f;
-
     // preferences
     public static final String SHARED_PREFS = "WiglePrefs";
     public static final String PREF_USERNAME = "username";
--- a/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/MappingFragment.java
+++ b/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/MappingFragment.java
@@ -163,7 +163,6 @@ public final class MappingFragment extends Fragment {
     private static final int OVERLAY_LIGHT = Color.parseColor("#F4D03F");
 
 
-
     /** Called when the activity is first created. */
     @Override
     public void onCreate(final Bundle savedInstanceState) {
@@ -183,7 +182,7 @@ public final class MappingFragment extends Fragment {
             locale = Locale.US;
         }
         numberFormat = NumberFormat.getNumberInstance(locale);
-        numberFormat.setMaximumFractionDigits(1);
+        numberFormat.setMaximumFractionDigits(2);
         // media volume
         getActivity().setVolumeControlStream(AudioManager.STREAM_MUSIC);
 
--- a/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/listener/GPSListener.java
+++ b/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/listener/GPSListener.java
@@ -27,13 +27,14 @@ import android.os.Bundle;
 import java.util.Collections;
 import java.util.Map;
 import java.util.TreeMap;
+import java.util.concurrent.atomic.AtomicLong;
 
 import androidx.core.content.ContextCompat;
 
 public class GPSListener implements Listener, LocationListener {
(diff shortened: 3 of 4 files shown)


wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/ListFragment.java
```diff
@@ -74,8 +74,6 @@ public final class ListFragment extends Fragment implements ApiListener, DialogL
     private static final int UPLOAD_DIALOG = 101;
     private static final int QUICK_PAUSE_DIALOG = 102;
 
-    public static final float MIN_DISTANCE_ACCURACY = 32f;
```
rksh, line 77: removing in favor of centralized MIN_ROUTE_LOCATION_PRECISION_METERS

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/MappingFragment.java
```diff
@@ -183,7 +182,7 @@ public void onCreate(final Bundle savedInstanceState) {
             locale = Locale.US;
         }
         numberFormat = NumberFormat.getNumberInstance(locale);
-        numberFormat.setMaximumFractionDigits(1);
+        numberFormat.setMaximumFractionDigits(2);
```
rksh, line 185: adding two decimals facilitates debugging as well as helping with the map view.

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/listener/GPSListener.java
```diff
@@ -27,13 +27,14 @@
...
 import java.util.TreeMap;
+import java.util.concurrent.atomic.AtomicLong;
 
 import androidx.core.content.ContextCompat;
 
 public class GPSListener implements Listener, LocationListener {
     public static final long GPS_TIMEOUT_DEFAULT = 15000L;
     public static final long NET_LOC_TIMEOUT_DEFAULT = 60000L;
-    public static final float LERP_MIN_THRESHOLD_METERS = 1.0f;
+    public static final float LERP_MIN_THRESHOLD_METERS = 20.0f;
```
rksh, line 37: am i nuts, or should we avoid linear interpolation unless the difference is a little larger than 1m

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/listener/GPSListener.java
```diff
@@ -254,35 +257,41 @@ else if ( newOK && NETWORK_PROVIDER.equals( newLocation.getProvider() ) ) {
...
-                            dist + prefs.getFloat( ListFragment.PREF_DISTANCE_RUN, 0f ) );
-                    edit.putFloat( ListFragment.PREF_DISTANCE_TOTAL,
-                            dist + prefs.getFloat( ListFragment.PREF_DISTANCE_TOTAL, 0f ) );
-                    edit.apply();
-                }
-                if ( dist > LERP_MIN_THRESHOLD_METERS && dbHelper != null) {
-                    if (dist > LERP_MAX_THRESHOLD_METERS) {
-                        Logging.warn("Diff is too large, not lerping. " + dist + " meters");
-                        dbHelper.clearPendingObservations();
+        if (location != null && location.getTime() != 0L &&
```
rksh, line 260: See DatabaseHelper.logRouteLocation (:1163) for almost-exactly-the-same-test

bobzilladev: This removes the check for GPS_PROVIDER, is that intentional?

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/listener/GPSListener.java
uhtu: why choose?

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/listener/GPSListener.java
uhtu: is that the source of the difference?

uhtu (approved):
anyhow, this should at least align the two.
👍

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/listener/GPSListener.java
rksh: _A_ source- but not the only one. You've seen the routes on your various devices, they don't look terrible is most cases.

rksh: There's probably a second PR where I try to eliminate jitter in both distance and route while holding still.

bobzilladev (approved):

PR #551, opened 2022-12-22
[FIX] first experiment in location -> distance

rksh:
Transition to GNSS provider is tickling unjustified route distance increases.
This prevents backtracking based on timestamp problems/race condition weirdness

this is still rough, setting do-not-merge

most of the change is just variable names changes.

### DIFF
--- a/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/listener/GNSSListener.java
+++ b/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/listener/GNSSListener.java
@@ -30,7 +30,6 @@ import java.util.TreeMap;
 import java.util.concurrent.atomic.AtomicLong;
 
 import androidx.core.content.ContextCompat;
-import androidx.core.location.LocationManagerCompat;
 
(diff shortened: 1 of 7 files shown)


wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/listener/GNSSListener.java
```diff
@@ -287,35 +286,38 @@ else if ( newOK && NETWORK_PROVIDER.equals( newLocation.getProvider() ) ) {
...
                     // set for next time; only update if this was a distance calc-event.
-                    prevGpsLocation = location;
+                    prevLocation = currentLocation;
                 }
-            } else {
+            } else if (prevLocation == null ||  prevLocation.getTime() < currentLocation.getTime()) {
                 // initialize previous location
-                prevGpsLocation = location;
+                prevLocation = currentLocation;
+            } else if (prevLocation != null){
```
rksh, line 300: yes, technically prevLocation shouldn't be able to be null here

PR #552, opened 2022-12-29
[ENH] Modernization

rksh:
working through legacy cruft and resource management problems that became apparently when upgrading to SDK 31

### DIFF
--- a/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/MainActivity.java
+++ b/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/MainActivity.java
@@ -26,7 +26,6 @@ import android.location.LocationManager;
 import android.location.LocationProvider;
 import android.media.AudioManager;
 import android.media.MediaPlayer;
-import android.media.MediaPlayer.OnErrorListener;
 import android.net.Uri;
 import android.net.wifi.WifiManager;
 import android.net.wifi.WifiManager.WifiLock;
@@ -34,6 +33,7 @@ import android.os.Build;
 import android.os.Bundle;
 import android.os.Handler;
 import android.os.IBinder;
+import android.os.Looper;
 import android.os.PowerManager;
 import android.preference.PreferenceManager;
 import android.provider.Settings;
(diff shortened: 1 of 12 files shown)


wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/ui/ThemeUtil.java
```diff
@@ -16,12 +18,24 @@
...
-            AppCompatDelegate.setDefaultNightMode(displayMode);
+            ExecutorService executor = Executors.newSingleThreadExecutor();
+            Handler handler = new Handler(Looper.getMainLooper());
+            executor.execute(() -> {
+                final int displayMode = prefs.getInt(ListFragment.PREF_DAYNIGHT_MODE, AppCompatDelegate.MODE_NIGHT_YES);
+                Logging.info("set theme called: " + displayMode);
+                handler.post(new Runnable() {
+                    @Override
+                    public void run() {
+                        AppCompatDelegate.setDefaultNightMode(displayMode);
```
bobzilladev, line 35: Does this really take long enough to need a background thread?

bobzilladev (approved):
Seems good! A lot of things going to executors, I think the DB helper will lock it properly, but some UI setting things as well which I'm surprised to see.

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/MainActivity.java
```diff
@@ -2705,31 +2635,31 @@ public void backgroundUploadFile(){
...
-            AlertDialog.Builder iseDlgBuilder = new AlertDialog.Builder(this);
-            iseDlgBuilder.setMessage(external?R.string.no_external_space_message:R.string.no_internal_space_message)
-                    .setTitle(external?R.string.no_external_space_title:R.string.no_internal_space_title)
-                    .setCancelable(true)
-            .setPositiveButton(R.string.ok, new DialogInterface.OnClickListener() {
-                @Override
-                public void onClick(DialogInterface dialog, int which) {
-                    dialog.dismiss();
+    public void checkStorage() {
+        ExecutorService executor = Executors.newSingleThreadExecutor();
```
uhtu, line 2639: we should consider (for future work) if we want to have a single executor for these uses.

uhtu (commented):
👍

PR #555, opened 2023-01-08
[FIX] Leak hunting

rksh:
- Getting rid of old leaky abstractions
- refactoring bluetooth receiver LE scanning to eliminate most common leak.
- removing some old conditional build checks
- leaning on try (with resource) when we can
- removing old workarounds in media player


wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/MainActivity.java
```diff
@@ -1097,140 +1094,6 @@ public static Locale getLocale(final Context context, final Configuration config
...
-    }
-
-    /**
-     * create a media player (trying several paths if available)
-     *
-     * @param context the context to use
-     * @param resid   the resource to use
-     * @return the media player for resid (or null if it wasn't creatable)
-     */
-    private static MediaPlayer createMp(final Context context, final int resid) {
```
uhtu, line 1199: RIP

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/listener/BluetoothReceiver.java
```diff
@@ -846,8 +719,176 @@ private Network addOrUpdateBt(final String bssid, final String ssid,
...
+                    final long gpsTimeout = prefs.getLong(PreferenceKeys.PREF_GPS_TIMEOUT, GNSSListener.GPS_TIMEOUT_DEFAULT);
+                    final long netLocTimeout = prefs.getLong(PreferenceKeys.PREF_NET_LOC_TIMEOUT, GNSSListener.NET_LOC_TIMEOUT_DEFAULT);
+                    gpsListener.checkLocationOK(gpsTimeout, netLocTimeout);
+                    location = gpsListener.getCurrentLocation();
+                } else {
+                    Logging.warn("Null gpsListener in LE Single Scan Result");
+                }
+            }
+
+            updater.get().handleLeScanResult(scanResult, location, false);
```
uhtu, line 774: this has need of that same weakref .get()-then-check-for-null

uhtu (approved):

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/MainActivity.java
rksh: it served valiantly

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/listener/BluetoothReceiver.java
rksh: righto, fixing.

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/MainActivity.java
bobzilladev: /salute

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/background/PooledQueryExecutor.java
```diff
@@ -0,0 +1,92 @@
...
+
+/**
+ * Execute raw parameterized DB queries off the main thread.
+ * Replacement for the old QueryThread system we used pre Android SDK 31 cutover.
+ * Maintains matching abstractions for ease of porting but uses an Executor instead of keeping a
+ * thread hot in the DatabaseHelper (and leaking).
+ * @author arkasha
+ */
+public class PooledQueryExecutor {
+    private static final ExecutorService executor = Executors.newFixedThreadPool(3);
```
bobzilladev, line 21: Part of the reason for that is single-threaded access, I assume that's still a requirement, and we still have appropriate locks in place, in which case no reason to have more than one thread in the thread pool? Or have some of these assumptions changed in the last decade?

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/listener/BluetoothReceiver.java
```diff
@@ -534,25 +404,30 @@ public void onReceive(final Context context, final Intent intent) {
                 prevBt = Collections.synchronizedSet(new HashSet(latestBt));
                 latestBt.clear();
 
-                ListFragment.lameStatic.currBt = prevBtle.size() + prevBt.size();
+                ListFragment.lameStatic.currBt = (null != scanCallback) ? scanCallback.getPrevBtLeSize():0 + prevBt.size();
```
bobzilladev, line 407: might want parens around the whole ternary? not sure if precedence rules will do what you intend there.

bobzilladev (approved):
some commentary, but pretty much lgtm

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/background/PooledQueryExecutor.java
rksh: I can switch this to single threaded executor  - glad you tipped me off. Re: thread safety, sounds like "it can be" https://developer.android.com/reference/android/database/sqlite/SQLiteDatabase

PR #556, opened 2023-01-20
check iv length, bump some versions

bobzilladev:
22 of 22 tests pass


rksh:
Seems cromulent - I didn't realize we needed to keep the maps ver in lockstep!

PR #557, opened 2023-01-20
move android-maps-utils to gradle

bobzilladev:
Increasing confidence this works at least as well as before. Only updating labels on map returning to idle after movement, batching handler messages for multiple networks at once, using looper so hopefully not leaking, and more dynamic cache size tuning.

15,151 deletes ftw.


wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/ListFragment.java
```diff
@@ -130,11 +130,11 @@ public static class LameStatic {
     static {
         final long maxMemory = Runtime.getRuntime().maxMemory();
         int cacheSize = 128;
-        if (maxMemory > 200000000L) {
-            cacheSize = 1024;
+        if (maxMemory > 400_000_000L) {
```
rksh, line 133: TIL

rksh (approved):
1. I love deleted code
2. We don't have an easy way to instrument this, but I'm betting this is a lot less battery intensive.

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/MapRender.java
```diff
@@ -329,15 +336,35 @@ private void sendUpdateNetwork(final String bssid) {
...
+                if (network != null) {
+                    networkRenderer.updateItem(network);
+                }
+            }
+
+            final ArrayList<String> bssids = message.getData().getStringArrayList(MESSAGE_BSSID_LIST);
+            if (bssids != null && !bssids.isEmpty()) {
+                Logging.info("bssids: " + bssids.size());
+                for (final String thisBssid : bssids) {
+                    final Network network = MainActivity.getNetworkCache().get(thisBssid);
```
uhtu, line 363: do these MainActivity getFoo() get hoisted? is it worth pulling them out of the loop?

PR #562, opened 2023-02-04
Executor fix

uhtu:
less service creation, tie into lifecycle

### DIFF
--- a/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/ListFragment.java
+++ b/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/ListFragment.java
@@ -121,6 +121,8 @@ public final class ListFragment extends Fragment implements ApiListener, DialogL
         public QueryArgs queryArgs;
         public ConcurrentLinkedHashMap<String,Network> networkCache;
         public OUI oui;
+
+        public ExecutorService executor;
     }
     public static final LameStatic lameStatic = new LameStatic();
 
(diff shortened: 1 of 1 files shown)


wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/ListFragment.java
```diff
@@ -196,7 +202,7 @@ private void setNetCountUI( final State state, final View view ) {
         if (view == null) {
             return;
         }
-        final ExecutorService executor = Executors.newFixedThreadPool(3);
+
```
bobzilladev, line 205: only thought is should we null check the executor calls now due to android wonkery

bobzilladev (approved):

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/ListFragment.java
uhtu: but muh lifecycles

bobzilladev: i...guess so. in a synchronized lazy initializer? to be as complicated as possible for something simple because we can't trust anything anywhere anytime? /me stares off into the distance, pushes "comment" button.

rksh: on not-trusting trust: mobile application frameworks

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/ListFragment.java
```diff
@@ -140,6 +140,11 @@ else if (maxMemory > 50_000_000L) {
         lameStatic.networkCache = new ConcurrentLinkedHashMap<>(cacheSize);
     }
 
+    /**
+     * for the doing of things
+     */
+    public ExecutorService executor;
```
rksh, line 146: not that we SHOULD have more than on ListFragment going at a time, but static would prevent concurrent access if we did.

rksh (approved):

PR #564, opened 2023-02-12
[ENH] Upload status <> human readable

rksh:
taking some feedback from previous code review, only to find a bug that needed fixing
 - Upload status now uses an enum.
 - Previous upload list row behavior hinged on human-readable strings, which was... broken.
 - Not sure if there's a more elegant way to translate to human readable but at least this isn't broken.

### DIFF
--- a/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/UploadsFragment.java
+++ b/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/UploadsFragment.java
@@ -62,18 +62,18 @@ public class UploadsFragment extends Fragment {
     private TextView queueDepth;
     private UploadsResponse latestResponse;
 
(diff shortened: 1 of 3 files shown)


uhtu (approved):

bobzilladev (approved):
Could bake the status map values directly into the enum, make sure they're always paired 1:1. But maybe you don't want to mingle that concept in there.

rksh:
we could move the _IDs_ of the strings in there, but we'd still need a way to populate the translated strings outside the model object (otherwise we're passing `Context` into either the constructor or the getter to get the results in the right language for the users).
Felt like  weird trade-off, but I could be convinced to add the R.strings.<id> values in as a second element to the Enum members if you think it's worthwhile/

bobzilladev:
blah, yeah that's pry too weird a task for the enum to be taking on.

PR #565, opened 2023-02-14
[ENH] Rankings display improvement.

rksh:
taking some old feedback, presenting more data.
also a small bug fix on change-mode


bobzilladev (approved):
gooey!

PR #566, opened 2023-02-15
[FIX] Preventing KML injection via crafted SSID

Krzysztofz01:
The current approach to filtering the SSID values before appending it to the KML file is to replace certain ranges of the charset. The usage of CDATA can filter most of the [XXE](https://portswigger.net/web-security/xxe) paylods which consist of `<`, `>` and `&` based tags. I found out, that there is still a posibility to inject a payload by escaping from the CDATA tag and then inserting XML/KML tags. To make things clear, this bug **IS NOT** a security vulnerability, but still a bug, which is able to generate a corrupted KML file, what can potentialy lead to *full*/*run* data loss. There is also a small propability that a corrupted KML file can cause problems with third-party applications (very old example: [CVE-2006-7157](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2006-7157)), but that's not really our concern.

The implementation still needs to be refined which characters should be additionally removed, and which implemented rules should be removed. The implementation itself is very simple and the best practice would be to use a *battle-tested* XML sanitization library.

PoC:
Test home-router `SSID` set to: `]]>test<![CDATA[`, after performing a scan and generating the KML file we can see the test network:
```xml
<!-- more data... -->
<Placemark>
    <name><![CDATA[]]>test<![CDATA[]]></name>
    <description><![CDATA[ more data... ]]></description>
    <styleUrl>more data...</styleUrl>
    <Point><coordinates>more data...</coordinates></Point>
</Placemark>
<!-- more data... -->
```
The fact that the `test` string is outside of the CDATA means that all type of XML/KML tags can be injected inside.

### DIFF
--- a/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/background/KmlWriter.java
+++ b/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/background/KmlWriter.java
@@ -3,6 +3,7 @@ package net.wigle.wigleandroid.background;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.OutputStream;
+import java.io.UnsupportedEncodingException;
 import java.text.SimpleDateFormat;
 import java.util.Date;
 import java.util.HashSet;
@@ -237,20 +238,17 @@ public class KmlWriter extends AbstractBackgroundTask {
                 style = "zeroConfidence";
             }
 
-            // not unicode. ha ha for them!
-            byte[] ssidFiltered = ssid.getBytes( MainActivity.ENCODING );
-            filterIllegalXml( ssidFiltered );
-            if (ssidFiltered.length == 0) {
-                ssidFiltered = NO_SSID.getBytes( MainActivity.ENCODING);
+            byte[] sanitizedSsid = sanitizeKmlString(ssid);
+            if (sanitizedSsid.length == 0) {
+                sanitizedSsid = NO_SSID.getBytes(MainActivity.ENCODING);
             }
-
 
             if (type.equals(NetworkType.WIFI)) {
 
                 final String encStatus = "Encryption: " + encryptionStringForCapabilities(capabilities) + "\n";
 
                 FileAccess.writeFos(fos, "<Placemark>\n<name><![CDATA[");
-                fos.write(ssidFiltered);
+                fos.write(sanitizedSsid);
                 FileAccess.writeFos(fos, "]]></name>\n");
                 FileAccess.writeFos(fos, "<description><![CDATA[Network ID: " + bssid + "\n"
                         + "Capabilities: " + capabilities + "\n" // ALIBI: not available on server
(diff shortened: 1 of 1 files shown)


rksh:
nice catch! thank you!

uhtu:
rather than escaping, to allow "legit" network ssids with attack characters in them, we could probably just modify the existing routine to replace-with-space on them?

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/background/KmlWriter.java
```diff
@@ -333,19 +331,38 @@ private static String encryptionStringForCapabilities(final String capabilities)
...
-                    (current >= 0x86 && current <= 0x9F)
-                    ) {
-                data[i] = ' ';
-            }
+    // Sanitize string by escaping reserved XML tokens and replace invalid charset ranges
+    private byte[] sanitizeKmlString(String value) throws UnsupportedEncodingException {
+        // Escape reserved KML/XML tokens in order to prevent XXE attacks. The text is inserted into
+        // a CDATA field, only "]]>" sanitization is required and the rest is redundant.
+        final String escapedValue = value
+                .replaceAll("]]>", "]]&gt;")
```
uhtu, line 339: each invocation of .replaceAll() is a new object allocation and Regex pattern compilation. every time. because java.

Krzysztofz01:
> rather than escaping, to allow "legit" network ssids with attack characters in them, we could probably just modify the existing routine to replace-with-space on them?

It can be approached in this way, but then we lose information about the *real* SSID.
It seems to me that only sanitizing CDATA escaping symbols can be addressed for now, as solving this problem prevents other options to perform a XXE. This will also solve the regex issue that [uhtu](https://github.com/uhtu) pointed out.

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/background/KmlWriter.java
```diff
@@ -333,29 +339,22 @@ private static String encryptionStringForCapabilities(final String capabilities)
...
-                .replaceAll("&", "&amp;")
-                .replaceAll(">", "&gt;")
-                .replaceAll("<", "&lt;")
-                .replaceAll("\"", "&quot;")
-                .replaceAll("'", "&apos;");
-
-        // "not unicode. ha ha for them!"
-        final byte[] escapedValueBytes = escapedValue.getBytes( MainActivity.ENCODING );
+        // Escape reserved KML/XML tokens in order to prevent XXE attacks.
+        value = KML_CDATA_OPEN_PATTERN.matcher(value).replaceAll("&lt;![CDATA[");
```
uhtu, line 343: .matcher() still makes a new object every time (even when it does not match) perhaps this can be runtime-simplified with a string.indexOf("<") || string.indexOf(">") which don't care about case, and then the expensive regex operation can be done only when it could possibly change the outcome?

Krzysztofz01:
Finally, as suggested by [uhtu](https://github.com/uhtu). The "greater than", "less than" symbols are simply replaced with the selected replacement symbol (space). Thus, the problem of additional allocations disappears (Regexes, etc.)

uhtu (approved):

PR #567, opened 2023-02-16
[ENH] queue pos in uploads dispy

rksh:
hey, that never worked?


bobzilladev (approved):
Working is the New Not Failing

PR #569, opened 2023-03-04
[FIX] Upload button leak

rksh:
By passing buttons references instead of general context.

Also:
- continuing to template text as possible
- cleaning up use of number formatting
- making main screen UI elements lay out cleanly RTL/LTR (some text fixes required yet)

### DIFF
--- a/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/ListFragment.java
+++ b/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/ListFragment.java
@@ -712,7 +712,7 @@ public final class ListFragment extends Fragment implements ApiListener, DialogL
             tv = view.findViewById( R.id.LocationTextView03 );
             final Activity a = getActivity();
             if (null != a) {
-                tv.setText(getString(R.string.list_speed) + " " + (location == null ? "" : metersPerSecondToSpeedString(state.numberFormat1, a, location.getSpeed())));
+                tv.setText(getString(R.string.list_speed, (location == null) ? "" : metersPerSecondToSpeedString(state.numberFormat1, a, location.getSpeed())));
             }
 
             TextView tv4 = view.findViewById( R.id.accuracy_text);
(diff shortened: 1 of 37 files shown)


wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/ListFragment.java
```diff
@@ -687,32 +705,30 @@ private void setLocationUI( final MainActivity main, final View view ) {
...
 
-            TextView tv4 = view.findViewById( R.id.LocationTextView04 );
-            TextView tv5 = view.findViewById( R.id.LocationTextView05 );
+            TextView tv4 = view.findViewById( R.id.accuracy_text);
+            TextView tv5 = view.findViewById( R.id.alt_text);
             if ( location == null ) {
                 tv4.setText( "" );
                 tv5.setText( "" );
             } else {
-                if (main != null) {
```
bobzilladev, line 707: Maybe add the main null check at the beginning of the method before its first dereference?

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/model/Network.java
```diff
@@ -276,7 +280,7 @@ public int hashCode() {
 
     @Override
     public boolean equals(final Object other) {
-        if (other != null && other instanceof Network) {
+        if (other instanceof Network) {
```
bobzilladev, line 283: why remove null check?

bobzilladev (approved):

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/model/Network.java
rksh: linter complained it could never be null. I know we've been over-cautious with this, but was this a load-bearing redundant null check?

bobzilladev: that might have been one of the ridiculous ones we saw in way off-brand chinese phones back in the day? i remember adding goofy nullchecks that should be impossible. maybe those days are behind us and this change is fine. we can always add back if it pops up on the reporting.

PR #573, opened 2023-03-19
[FIX] Race condition fixes

rksh:
- make sure mainAcitivity is still defined before using it to get the template string.
- make sure DB isn't closed before writing a route point.


uhtu (approved):
fascinating!

bobzilladev (approved):
this is the way

PR #585, opened 2023-04-16
[FIX] responding to crash reports.

rksh:
Some of these null checks are deemed impossible by the linter, but here we are with reports that they've happened


bobzilladev (approved):
hello nullness my old friend, i've come to code with you again

PR #586, opened 2023-04-25
Only move camera if there were results

bobzilladev:
Somehow had an IllegalStateException get out of that catch. Might as well check that there is anything to move to and ignore the move entirely.


rksh (approved):
I mean, if this doesn't fix it, things are getting destroyed in-flight.

PR #592, opened 2023-05-09
[ENH] creating 'I'm loading' animations for network based uploads- and rank-view activities

rksh:
explain that these two blank screens are trying to do something.


wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/ui/ProgressThrobberFragment.java
```diff
@@ -0,0 +1,53 @@
...
+
+import net.wigle.wigleandroid.R;
+import net.wigle.wigleandroid.util.Logging;
+
+/**
+ * Loading "throbber" image parent class for network-based fragments.
+ */
+public abstract class ProgressThrobberFragment extends Fragment {
+    protected ImageView loadingImage;
+    protected boolean animating = false;
```
bobzilladev, line 18: Does this need to be Atomic? 🤯

bobzilladev (approved):

bobzilladev:
A throbbing good time!

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/ui/ProgressThrobberFragment.java
rksh: I *think* this is only manipulated with the GUI thread (but check my work)
onCreateView
RequestCompletedListener.onTaskCompleted (see uses in WiGLEApiManager - should always post to main GUI thread)

bobzilladev: yay for gooey, ok dokey

PR #594, opened 2023-05-29
[ENH] Login checks and failure handling

rksh:
making the client behave in the event of various network and auth failures


wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/UserStatsFragment.java
```diff
@@ -102,7 +109,17 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container, Bundle sa
         final ScrollView scrollView = (ScrollView) inflater.inflate(R.layout.userstats, container, false);
         MainActivity.State s = MainActivity.getStaticState();
         if (s != null) {
-            s.apiManager.getUserStats(new RequestCompletedListener<UserStats, JSONObject>() {
+            s.apiManager.getUserStats(new AuthenticatedRequestCompletedListener<UserStats, JSONObject>() {
+                @Override
+                public void onAuthenticationRequired() {
```
bobzilladev, line 114: Wonder if this could be in a super class to not have to put it in multiple places.

bobzilladev (approved):
When a problem comes along, you must throb it.

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/UserStatsFragment.java
bobzilladev: May have missed this comment, is there a way to not specify this 4 different times using some class structure?

bobzilladev: not a hard request, but just wanted your take on it, might make it all easier to deal with, or it might be impossible because android.

PR #596, opened 2023-05-30
[FIX] Refactor login dialog, lint/cleanup

rksh:
Able to condense all but one of the authentication dialog calls.
Also sacrificing unused includes and unreachable version checks to the lint gods./
The remaining `AppCompatActivity` superclass used in `DBResultActivity` doesn't appear easily reconcilable with `FragmentActivity` (we actually use a lot of the functionality in the child class)


bobzilladev (approved):
wooty

PR #598, opened 2023-06-02
[ENH] Continuing strings cleanup

rksh:
- making the network UI suck less
- templating/simplifying strings
- general XML view cleanup


bobzilladev (approved):
ooey gui

PR #600, opened 2023-06-06
[ENH] XML layout fixes and an upload endpoint patch

rksh:
cleaning up a bunch of outdated stuff
compensating for the fact that "STATS" status doesn't list new+location.


bobzilladev (approved):
lay em out, Kris

PR #601, opened 2023-06-06
[ENH] establishing a post-support "Oldroid" branch

rksh:
One of the reasons EoL devices (notably android 4.x) cannot maintain functionality is the lack of https support - this PR will keep things going for devices with working TLS1.3 devices where possible, for users willing to sideload.

This won't necessarily work for FOSS-only devices.

Also updates oui definitions.


uhtu (approved):
👍

bobzilladev (approved):
wooty, we can bake a release sdk for this branch

PR #602, opened 2023-06-12
[FIX] ISE in DBResults

rksh:
trying to remedy the 'IllegalStateException: no included points for empty results' fed to the map-ready call.


bobzilladev (approved):
Thanks! I had this one in my todo list.

PR #606, opened 2023-06-16
[ENH] removing dead code

rksh:
and fixing up messy modular composition which is no longer needed.
Also some stability improvements from 2.75 bugs

### DIFF
--- a/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/SettingsFragment.java
+++ b/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/SettingsFragment.java
@@ -283,7 +283,7 @@ public final class SettingsFragment extends Fragment implements DialogListener {
         final View authUserLayout = view.findViewById(R.id.show_authuser_label);
         final EditText passEdit = view.findViewById(R.id.edit_password);
         final View passEditLayout = view.findViewById(R.id.edit_password_label);
-        final CheckBox showPass = view.findViewById(R.id.showpassword);
+        final CheckBox showPassword = view.findViewById(R.id.showpassword);
         final String authToken = prefs.getString(PreferenceKeys.PREF_TOKEN, "");
         final Button deauthButton = view.findViewById(R.id.deauthorize_client);
         final Button authButton = view.findViewById(R.id.authorize_client);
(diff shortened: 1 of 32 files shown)


wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/SettingsFragment.java
```diff
@@ -409,26 +409,11 @@ public void onTaskFailed(int status, JSONObject error) {
...
-                register.setText(Html.fromHtml(registerBlurb,
-                        Html.FROM_HTML_MODE_LEGACY));
-            } else {
-                register.setText(Html.fromHtml(registerBlurb));
-            }
+            register.setText(Html.fromHtml(getString(R.string.registration_html_prompt),
+                    Html.FROM_HTML_MODE_LEGACY));
         } catch (Exception ex) {
-            register.setText(registerString + " @WiGLE.net");
+            Logging.error("unable to create registration prompt from HTML");
```
bobzilladev, line 416: Should we keep setting something in the UI for the user to try and do something with here? Or is fromHtml really unlikely to fail?

wiglewifiwardriving/src/main/res/values-ja-rJP/strings.xml
```diff
@@ -444,4 +444,8 @@
     <string name="login_title">ログインが必要です</string>
     <string name="login_required">これを行うにはログインする必要があります。必要に応じて、ログインするかアカウントを作成してください。</string>
     <string name="login">ログイン</string>
+    <string name="registration_html_prompt">&lt;a href=\'net.wigle.wigleandroid.register://register\'>WiGLE.net に登録する&lt;/a&gt;か、&lt;a href=\'net.wigle.wigleandroid://activate\'&gt;QR コードで今すぐアクティベートしてください&lt;/a&gt;</string>
```
bobzilladev, line 447: This file doesn't delete `register` or `activate`

bobzilladev (changes requested):

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/SettingsFragment.java
rksh: Good question - it's just the prompt to register, and if we can't turn it into links, we'd... have to add "register for an account at WiGLE.net" in 40 languages with no link, maybe? Regardless, it's pretty unlikely to fail, and failing quietly doesn't change much other than access to the activate activity.

bobzilladev (approved):

PR #607, opened 2023-07-12
condense app_name to required 30 chars

bobzilladev:


rksh (approved):
admittedly, my skills in these languages are limited, but lgtm.
Should we be renaming the app to WiGLE Wireless or just WiGLE Wardriving, since it's not Wiffies-specific?

PR #610, opened 2023-08-07
[FIX] fixing some NPE risks

rksh:
responding to some user reported issues.


bobzilladev (approved):
someday we're going to need to Option this, bleh

PR #612, opened 2023-08-10
obey our own file spec times

bobzilladev:


rksh:
Per investigation issues reported by kgrenades during the WWWD

PR #617, opened 2023-08-18
[ENH] layout cleanup.

rksh:
Now that we're free of some of the older APIs/quirks, we can fix the layout in our list view. This partitions the elements of a row into their own layout cells, rather than pre-composing a long string and putting it into a single second-row cell.

I would prefer that this work in even columss, but the length of WiFi capabilities (building from the "end" of the row) and Cell NetIDs (building from the start) makes  columnar spacing too challenging.

```
+--------------------------------------------------- +
| 310260_14553_21139798   ...                        |
+--------------------------------------------------- +
```
vs wifi, which fills the line completely (and sometimes further)
```
+--------------------------------------------------- +
| 88:00:00:12:00:00  5888MHz  [WPA2][RSN][ESS][WPS]  |   
+--------------------------------------------------- +

```

Channels have become almost meaningless with so many 802.11<whatever> standards defining overlapping numbers, so we're squeezing in center frequency in MHz for WiFi- this is quite tight but works even on narrow aspect-ratio devices in testing.

making signal numbers fit under icons a little better.
doing less text-character formatting, more spacing/color-differentiation

Partitioning the layout as follows:
```
+--------------------------------------------------------- +
| ICO | NAME           | MFGR                  | time      |
+--------------------------------------------------------- +
| sig |  NETID             | FREQ |           CAPABILITIES |
+--------------------------------------------------------- +
```

including `tools:` namepsace sample text to help understand layout

### DIFF
--- a/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/model/Network.java
+++ b/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/model/Network.java
@@ -230,18 +230,11 @@ public final class Network implements ClusterItem {
 
     public String getDetail() {
         if ( detail == null ) {
-            final Integer chan = channel != null ? channel : frequency;
             final StringBuilder detailBuild = new StringBuilder( 40 );
-            detailBuild.append( BAR_STRING ).append( bssid );
-            detailBuild.append( DASH_STRING );
-            if ( NetworkType.WIFI.equals(type) ) {
-                detailBuild.append( chan );
+            if (!NetworkType.WIFI.equals(type)) {
+                detailBuild.append(type).append(BAR_STRING);
             }
-            else {
-                detailBuild.append( type );
-            }
-
-            detailBuild.append( DASH_STRING ).append( getShowCapabilities() );
+            detailBuild.append( getShowCapabilities() );
             detail = detailBuild.toString();
         }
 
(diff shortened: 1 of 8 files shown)


bobzilladev:
Runs fine for me! Couple aesthetic comments:

- The frequency (or ble/bt) indicator looks to be slightly smaller font than the bssid and capabilities, and is vertically oriented to the top of that row, so it looks like it's floating higher above the bottom line than the other text. I imagine it would look more consistent if that can somehow be vertically pushed to the bottom, so it has the same spacing from bottom of text to the line as the bssid and capabilities.
- Wondering if the capabilities should also be the darker grey like freq and time. Maybe not. Just seems like there's a bunch of different shades going on now, where before it was more uniform.

rksh:
> Runs fine for me! Couple aesthetic comments:
> 
> * The frequency (or ble/bt) indicator looks to be slightly smaller font than the bssid and capabilities, and is vertically oriented to the top of that row, so it looks like it's floating higher above the bottom line than the other text. I imagine it would look more consistent if that can somehow be vertically pushed to the bottom, so it has the same spacing from bottom of text to the line as the bssid and capabilities.

Agreed- fixed!

> * Wondering if the capabilities should also be the darker grey like freq and time. Maybe not. Just seems like there's a bunch of different shades going on now, where before it was more uniform.

This is somewhat intentional because of the shared-row-layout issue above. If strings overrun/get too long, the different shades help differentiate.

bobzilladev (approved):

Issue #615, opened 2023-08-15
Record Bluetooth Manufacturer-specific data


XenoKovah:
This is a request to start collecting additional information from Bluetooth Low Energy Advertisements & Bluetooth Classic Extended Inquiry Responses, in order to better determine which company a device is associated with.

One of the types of information that can be advertised by a device is "Manufacturer-specific" data (type 0xff, https://developer.android.com/reference/android/bluetooth/le/ScanRecord#DATA_TYPE_MANUFACTURER_SPECIFIC_DATA) which is *supposed to be* a 16 bit company ID followed by arbitrary-length service-specific data.

The company IDs can be looked up here: https://bitbucket.org/bluetooth-SIG/public/src/main/assigned_numbers/company_identifiers/company_identifiers.yaml (Note: this is not the same as the UUID16 company IDs from ticket #614 )

Within my data, this data type is by far the most common type of advertised data. However, because the company IDs extractable from this field are less accurate than those from the other UUID16 entries (like type 0x2 and 0x3 mentioned in ticket #614 ), it should perhaps be done after data extraction for those types.

Example with Company ID:
<img width="1519" alt="Pasted Graphic 16" src="https://github.com/wiglenet/wigle-wifi-wardriving/assets/92380610/7e5051e0-3761-47a5-a074-fd280cbddffb">

---
***NOTE:** This type of information can also appear in Bluetooth Classic Extended Inquiry Responses.*

Example:
<img width="1524" alt="Pasted Graphic 15" src="https://github.com/wiglenet/wigle-wifi-wardriving/assets/92380610/f265a9a0-5f91-4606-96bb-0d555852a215">

Note that sometimes the vendor IDs may be endian-swapped. For instance in the below data, it says it's an iPhone but the company ID is 0x4C00. The correct Apple ID (from [here](https://bitbucket.org/bluetooth-SIG/public/src/main/company_identifiers/company_identifiers.yaml)) is 0x004C. (In reality I wasn't able to find a location in the spec that specifically said that the 16-bit company ID should be little-endian, so vendors may have changed their endianness over time.)
<img width="1522" alt="Pasted Graphic 13" src="https://github.com/wiglenet/wigle-wifi-wardriving/assets/92380610/16c62a0e-7816-4ede-988e-16856dfa483f">

And sometimes they just don't seem to correspond to anything, and my suspicion is that vendors are just skipping the company ID and putting arbitrary data into the advertised type.
[...]

Issue #614, opened 2023-08-15
[Feature request] Record Bluetooth UUID16s


XenoKovah:
This is a request to start collecting additional information from Bluetooth Low Energy Advertisements & Bluetooth Classic / EDR Extended Inquiry Responses, in order to better determine which company a device is associated with.

One of the types of information that can be advertised by a device is a UUID16. These can appear in either "complete" (type 3, https://developer.android.com/reference/android/bluetooth/le/ScanRecord#DATA_TYPE_SERVICE_UUIDS_16_BIT_COMPLETE) or "incomplete" / "partial"(type 2, https://developer.android.com/reference/android/bluetooth/le/ScanRecord#DATA_TYPE_SERVICE_UUIDS_16_BIT_PARTIAL) arrays of uint16s.

While the UUIDs are called "Service Class" UUIDs, usually when they are lower numbers (0x1XXX) and increasing they correspond to services (which can be looked up with https://bitbucket.org/bluetooth-SIG/public/src/main/assigned_numbers/assigned_numbers/uuids/service_uuids.yaml), but when they are higher numbers (0xFEXX) and decreasing they correspond to company IDs (which can be looked up with https://bitbucket.org/bluetooth-SIG/public/src/main/assigned_numbers/assigned_numbers/uuids/member_uuids.yaml).

The below screenshots are taken from looking at HCI logs, collected by the Linux btmon tool, in Wireshark (not raw packet captures.) However, the data is structured equivalently to what would be within a packet.

Here is an example of the Fitbit company ID UUID16 (0xfd63) appearing as part of an **"incomplete" (type 2)** list:
![Pasted Graphic](https://github.com/wiglenet/wigle-wifi-wardriving/assets/92380610/7ebfc95b-3359-4689-bc12-cd221c86034d)

Here is an example of a service (0x1122 - Basic Printing) appearing in an "incomplete" (type 2) list: (for a Tesla, based on the name (^S[0-9a-f]{16}C$)- https://trifinite.org/Downloads/20220916_tempa_presentation_sec-t_public.pdf)
<img width="1525" alt="Pasted Graphic 3" src="https://github.com/wiglenet/wigle-wifi-wardriving/assets/92380610/c1f4ddb1-19b4-46f8-843b-2ba7a1d3ab58">

Here is an example of the Logitech ID appearing as part of a **"complete" (type 3)** list:
[...]

PR #620, opened 2023-08-21
[ENH] BT ID experiment

rksh:
this is NOT production ready, but this is a stab at decoding BT vendor IDs per #614 #615 . Android does *some* of the heavy lifting for us on mfgr ID cleanup, although it's unclear how reliable this is (and appears to be left to Android vendors?)
Open to feedback about how to do this in a way that would be suitable for production integration (and attendant efficient storage for the mfgr ID and service-uuids on device.)

NB: since MAC address randomization renders individual MACs alias-able, the primary key of any storage for these things would need to be a compound key of both MAC and date and or location.


rksh:
Note that because the mfgr payloads contain information that serves as a fingerprint in some cases, we would never want to allow these payloads onto the server.

wiglewifiwardriving/build.gradle
```diff
@@ -131,3 +134,21 @@ task createMccMncDB (dependsOn: 'downloadFile') {
...
+    download {
+        //TODO: this is a specific version - we want latest (1/2)
+        src 'https://bitbucket.org/bluetooth-SIG/public/raw/1ac368526f665502b73c68c9cd999cb03882e5e5/assigned_numbers/uuids/member_uuids.yaml'
+        dest outputBtMemberUuidFile
+        overwrite true
+    }
+    download {
+        //TODO: this is a specific version - we want latest (2/2)
+        src 'https://bitbucket.org/bluetooth-SIG/public/raw/1ac368526f665502b73c68c9cd999cb03882e5e5/assigned_numbers/company_identifiers/company_identifiers.yaml'
+        dest outputBtCoIdUuidFile
```
rksh, line 150: how to download the HEAD of a bitbucket link in raw, rather than a specific rev.?

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/listener/BluetoothReceiver.java
```diff
@@ -201,29 +207,56 @@ public void handleLeScanResult(final ScanResult scanResult, Location location, f
...
+                        uuid16Services.add(u.getUuid().toString());
+                    }
+                }
+
+                Integer mfgrKey = null;
+                if (null == uuid16Services && scanRecord.getManufacturerSpecificData() != null) {
+                    SparseArray<byte[]> bytesArray= scanRecord.getManufacturerSpecificData();
+                    //String hexDump = "";
+                    for(int i = 0; i < bytesArray.size(); i++) {
+                        mfgrKey = bytesArray.keyAt(i);
```
rksh, line 239: Android is parsing these out for us (great!) and they map - but unclear whether it's good at it. Casual testing looks good. Avoid storing the whole payload due to privacy concerns.

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/model/Network.java
```diff
@@ -46,11 +53,14 @@ public final class Network implements ClusterItem {
     private LatLng geoPoint;
     private boolean isNew;
 
+    private List<String> bleServiceUuids;
+    private Integer bleMfgrId;
+    private String bleMfgr;
```
rksh, line 58: this is really non-ideal

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/model/Network.java
```diff
@@ -208,6 +237,26 @@ public void setSsid(final String ssid) {
...
+            bleMfgr = lookupMfgrByServiceUuid(bleServiceUuids.get(0));
+        }
+    }
+
+    public void addBleMfgrId(Integer id) {
+        bleMfgrId = id;
+        if (bleMfgrId != null) {
+            bleMfgr = lookupMfgrByMfgrId(id);
+        }
+    }
```
rksh, line 259: these two methods will fight over the mfgr string. Not ideal, not reliable, esp considering the caveats inherent in each one.

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/model/Network.java
rksh: I guess storing the source data is preferable. The string we're holding onto will be a combination of the two, but in practice, I'm not seeing a lot of overlap, and never a disagreement.

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/listener/BluetoothReceiver.java
rksh: In practice, Android appears to be VERY good at parsing this out. assume there's a byte order flag it knows to use elsewhere in the payload. This virtually never fails to ID a BT device, and always correlates with other factors (device name, actual device identification) in testing.

wiglewifiwardriving/src/main/res/layout-v29/row.xml
```diff
@@ -48,20 +49,24 @@
 				android:ellipsize="end"
 				android:maxLines="1"
 				style="@style/ListOui"
-				tools:text="WiGLE NIC, Inc"
+				android:layout_weight="1"
+				android:gravity="end"
+				tools:text="WiGLE, a subsidiary of Very, Very Long Name, Inc"
```
bobzilladev, line 54: Was this a test name that should go back to what it was?

wiglewifiwardriving/src/main/res/layout/row.xml
```diff
@@ -48,13 +49,16 @@
 				android:ellipsize="end"
 				android:maxLines="1"
 				style="@style/ListOui"
-				tools:text="WiGLE NIC, Inc"
+				android:layout_weight="1"
+				android:gravity="end"
+				tools:text="WiGLE, a subsidiary of Very, Very Long Name, Inc"
```
bobzilladev, line 54: same long name revert?

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/listener/BluetoothReceiver.java
```diff
@@ -201,29 +207,56 @@ public void handleLeScanResult(final ScanResult scanResult, Location location, f
...
+                        //final MainActivity ma = MainActivity.getMainActivity();
+                        //if (null != ma) {
+                        //    final String mfgrName = ma.getBleMfgr(key);
+                        //    Logging.error("got: "+mfgrName+" for "+key+" net "+bssid);
+                            //return "~"+mfgrName;
+                        //} else {
+                        //    Logging.error("null mfgr name: "+key);
+                        //}
+                    }
+                    //Logging.error("Need to try: "+hexDump);
```
bobzilladev, line 250: Should this commented-out debug printing be removed?

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/listener/BluetoothReceiver.java
```diff
@@ -607,13 +641,14 @@ private Network addOrUpdateBt(final String bssid, final String ssid,
         boolean btTypeUpdate = false;
         if (network == null) {
             //DEBUG: MainActivity.info("new BT net: "+bssid + "(new: "+newForRun+")");
-            network = new Network(bssid, ssid, frequency, capabilities, strength, type);
+            //ALIBI: using frequency to hold deviceType
+            network = new Network(bssid, ssid, deviceType, capabilities, strength, type, uuid16Services, mfgrId);
```
bobzilladev, line 645: not sure this optimization of devicetype in frequency field is worth the unexpectedness of it, especially when we're already adding more fields to Network. like we'd have to special case this in the csv writer to not put devicetype in frequency field. maybe network could have a "subtype" or something to be general?

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/model/Network.java
```diff
@@ -190,13 +216,16 @@ public Integer getChannel() {
...
+        this.capabilities = capabilities;
+    }
 
     // Overloading for *FCN in GSM-derived networks for now. a subclass is probably more correct.
     public void setFrequency( final int frequency) {
         this.frequency = frequency;
         if (NetworkType.WIFI.equals(this.type)) {
             this.channel = channelForWiFiFrequencyMhz(frequency);
-        } else if (frequency != 0 && frequency != Integer.MAX_VALUE) {
+        } else if (!NetworkType.BLE.equals(this.type) && !NetworkType.BT.equals(this.type) && frequency != 0 && frequency != Integer.MAX_VALUE) {
```
bobzilladev, line 228: More special casing because of the datatype optimization, would avoid :)

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/model/Network.java
bobzilladev: Could the latter only set if `bleServiceUuids` is null or empty? assuming we want to give priority to the uuid-set one?

bobzilladev (changes requested):
This turned out to be pretty fun! A few comments.

wiglewifiwardriving/src/main/res/layout-v29/row.xml
rksh: oh, this is me learning to use the "tools:text" - they're layout placeholders for use during XML layout design - in this case, since this field is the one that should get truncated and "ellipsized" (and not the surrounding fields) adding the design-time placeholders let me make sure my layout is working as intended.
I actually learned to use these because of your feedback about how the MHz were laying out.

wiglewifiwardriving/src/main/res/layout/row.xml
rksh: same as previous.

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/model/Network.java
rksh: I'm not at all clear they're "better" - I think Xeno preferred service UUIDs because they were easier to deserialize, but I think Android successfully gets byte order right, so the Mfgr ID is quite realiable.

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/listener/BluetoothReceiver.java
rksh: this was a legacy hack from a few years ago, when we'd hoped to avoid modifying the `Network` object. I'd prefer to do this in an independent PR (I just renamed the variable to make the hack stand out)

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/model/Network.java
bobzilladev: ah, was hoping we'd always want to skew in one direction, could then make it bias in one way for deterministic results.

wiglewifiwardriving/src/main/res/layout-v29/row.xml
bobzilladev: ah, TIL

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/listener/BluetoothReceiver.java
bobzilladev: "We can't do this! It's horrible"
"Oh my sweet summer child, we've been doing this since before you were born"
*shoves kid out the window*

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/listener/BluetoothReceiver.java
rksh: aye, fixed!

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/model/Network.java
rksh: so, there can be more than one service UUID, and Xeno contends the order _isn't_ significant - so you might get a Bose-Google-Amazon as in the discussion for https://github.com/wiglenet/wigle-wifi-wardriving/issues/614 - the Mfgr bytes might actually be more reliable/meaningful.

bobzilladev: Do we want to bias the other way, where a uuid will only set on the object if mfgr is null, to allow mfgr bytes to always take precedence? I'd love to have determinism 😄

Issue #619, opened 2023-08-21
Dark mode, barely readable text


pschonmann:
When dark mode is activated, text is hard to read.
![Screenshot_20230821-092624](https://github.com/wiglenet/wigle-wifi-wardriving/assets/6333844/aa2a4edd-b0b0-4145-930a-5fb8e7d3cd89)

rksh:
This is a common issue with non-flagship Samsung phones post-orientation change - please annotation with the phone and OS version.
You can prevent this by
1. switching to light mode (ick, I know)
2. locking orientation so the rotation doesn't screw up the list view (other OS events might break this)
3. try experimenting with the relationship of Settings display mode and day-night settings to the OS.

We're aware of it, but haven't yet been able to override Samsung's weird interface hacks in "night mode"

rksh:
this is a recurrence of https://github.com/wiglenet/wigle-wifi-wardriving/issues/475

rksh:
one other possible solution (don't know why this sometimes works):
1. Set the app to "Light" mode in Settings. 
2. Leave settings, return to list view and rotate the screen a few times to make sure it remains readable after refresh.
3. Exit the app.
4. Set the app to "Dark" mode in Settings.
5. Leave settings, return to list view and rotate the screen a few times to make sure it remains readable after refresh.
6. Exit the app
7. Relaunch.

Try rotating the screen, see if the app maintains correct text rendering.

I'm not sure where in Samsung's stack this stuff is getting cached or broken, but this does clear the bug in testing with an A10e.

rksh:
marking fixed, slated for 2.79

PR #621, opened 2023-08-22
[FIX] Network list theme loss

rksh:
hacking a theme-change aware network list view - answering #619
This thing keeps popping back up - mostly SMSN users.

marked Do Not Merge pending further testing


bobzilladev (approved):
seems ok, as long as it's actually legal to swap that on the fly. probably only device testing will tell.

rksh:
testing looks good on:
Samsing Z Flip3
Pixel 5
Samsung a10e

PR #622, opened 2023-08-24
users love approving things, give them POST_NOTIFICATIONS for api33

bobzilladev:
https://developer.android.com/reference/android/Manifest.permission#POST_NOTIFICATIONS

This seems to make our notifications work again on api33. Irony that they were so strict about having services have notifications, but now make you opt in to showing them.


rksh (approved):
Does what it says on the tin (and what google tells you to do)

PR #627, opened 2023-09-07
[ENH] layout design helpers and two small fixes

rksh:
adding some design-time hints, and using them to "ellipsize" two fields which frequently cause runtime layout changes


bobzilladev (approved):
fancy

PR #629, opened 2023-09-07
[FIX] notification annoyances

rksh:
some conjunction of target SDK and notification changes was playing a lot of noise -> one star reviews.


bobzilladev (approved):
DWIM

Issue #333, opened 2019-05-03
Add bluetooth count to speech


goldfndr:
Currently (2.42), there's no checkbox in the "Speech Settings" for New bluetooth to DB, and I'm not hearing counts.

bobzilladev:
Will be in next release: https://github.com/wiglenet/wigle-wifi-wardriving/pull/632

PR #632, opened 2023-09-10
add bluetooth text to speech

bobzilladev:
For #333


rksh (approved):
👍

wiglewifiwardriving/src/main/res/values-iw/strings.xml
uhtu, line 1: i appreciate that this is android's issue not ours, but it is odd that iw and he are both required?

uhtu (commented):
👍

PR #636, opened 2023-09-18
[ENH] adding and updating distance to main list view

rksh:


bobzilladev (approved):
I love that we store distance in prefs, like gangsters

rksh:
We’ve got to research our static strategy, too!

PR #644, opened 2023-10-18
[FIX] take YAML load out of the way of startup

rksh:
startup reliability diminished after this change and the new version bump. let's background these loads. Pathology: app would start, but not start scanning, if the user locks the screen, scanning stays stopped until they unlock again. Moving things out of the GUI thread seems like a logical first step in tracing that, and this probably doesn't belong there anyway!

Also minor linting:
- unwrapping an always-true version check
- `@NotNull`ing to match overridden method


bobzilladev (approved):
Backgrounding is betterer. Could we compile in with Properties like the other lookups?

PR #648, opened 2023-10-27
Add [HS2.0] to capabilities

mark-grayson:
Add "[HS2.0]" to list of capabilities for Passpoint networks to allow easy indication of which networks may have RCOIs configured


bobzilladev:
@rksh would we rather do this with more iconography? I think I'm against altering the raw data we get from the OS.

mark-grayson:
one of the challenges is that Passpoint is orthogonal to WPA2/3, so there would need to be something like HS2/WPAII and HS2/WPAIII, e.g., 
image

rksh:
https://github.com/wiglenet/wigle-wifi-wardriving/pull/757 moving us closer to implementing something like this, although I believe an overlay is more appropriate.

PR #651, opened 2023-11-04
turn off notification sound

bobzilladev:


bobzilladev:
Updated to just `channel.setSound(null,null)`

bobzilladev:
Now checks for `samsung` to get HIGH

bobzilladev:
now checks for `google` to _not_ get HIGH

bobzilladev:
ok, now only does `setSound(null,null)`

rksh (approved):

PR #652, opened 2023-11-06
[ENH] notification priority fix

rksh:
testing confirms bob's assertion that DEFAULT is sufficient with a new or replaced notification channel ID


bobzilladev (approved):
woot!

PR #653, opened 2023-11-28
[FIX] search fix

rksh:
remedies production npe

### DIFF
--- a/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/SearchFragment.java
+++ b/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/SearchFragment.java
@@ -338,7 +338,7 @@ public class SearchFragment extends Fragment {
                     Geocoder geocoder = new Geocoder(context);
                     try {
                         List<Address> addressList = geocoder.getFromLocationName(location, 1);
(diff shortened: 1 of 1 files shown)


wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/SearchFragment.java
```diff
@@ -334,21 +334,23 @@ private void setupAddressSearch(final Context context, final View view) {
...
             public boolean onQueryTextSubmit(String query) {
                 String location = searchView.getQuery().toString();
-                List<Address> addressList = null;
                 if (location != null || location.equals("")) {
                     Geocoder geocoder = new Geocoder(context);
                     try {
-                        addressList = geocoder.getFromLocationName(location, 1);
+                        List<Address> addressList = geocoder.getFromLocationName(location, 1);
+                        if (null != addressList) {
+                            Address address = addressList.get(0); // ALIBI: taking the first choice. We could also offer the choices in a drop-down.
```
bobzilladev, line 342: saw an IndexOutOfBoundsException as well, probably want an isEmpty check on the list too?

bobzilladev (approved):

PR #656, opened 2023-12-14
[FIX] Linting and cleanup

rksh:


wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/background/ObservationUploader.java
```diff
@@ -150,7 +150,6 @@ private void doRun() throws InterruptedException, WiGLEAuthException {
      * override base startDownload - but instead perform an upload
      * TODO: a misnomer, really
      * @param fragment the fragment from which the upload was started
-     * @throws WiGLEAuthException
```
bobzilladev, line 153: Doesn't really matter, but it does still throw this.

bobzilladev (approved):
woot

PR #657, opened 2023-12-17
[FIX] Build action fixes

rksh:
absolute can of worms, attempting to get github build action working:
- getting things working so we can use gradle 8
- upgrade download task library changes method signature from `download {...}` to `download.run {...}`
- testing manual gradlew points out we need to increase build memory size
-` jcenter()` is deprecated and would hold us back, replacing with `mavenCentral()` and...
- `MaskedTextEdit` is basically abandoned, won't move from jcenter to mavenCentral - so we're resorting to local flat lib files
- create a `lib` dir and add the `MaskedTextEdit` items to it
- Our certificate transparency library has switched maintainers and versions in mavenCentral, so port that
- switch from manual paths to `layout.xx` `file` and `dir` - gradle convention (and certainly removes a security risk)
- reconfigure out staging dir for the MCC/MNC download/sqlite alter to fit inside the `layout.projectDir` hierarchy
- change to the `tasks.register` task definition style, make the download/DB synthesis tasks preconditions for build
- add some messages to the gradle download tasks so we know what's working/not


bobzilladev (approved):
One comment

PR #658, opened 2023-12-19
[FIX] Massive linting

rksh:
We finally got below the limit to get all our build warnings in a report while working to get the build action working!
This fixes a bunch of the warnings.
- fixed issues that would prevent us from migrating to gradle 8
- fixing layout overlaps
- got rid of some spurious warnings
- handling `Locale`s for string geneeration
- wrapping build-specific constants with SDK version checks
- getting rid of no-longer-needed SDK version checks
- mass-moving-about of assets which were mis-located
- elimination of unused resources
- creation of monochrome launcher icon packs
- resizing of launcher icon to avoid bleed
- fixing some layout ID conflicts
- ellipsizing some raw `...`s


wiglewifiwardriving/src/main/java/net/wigle/m8b/geodesy/mgrs.java
```diff
@@ -74,7 +76,8 @@ public static mgrs fromUtm(utm u){
 	return m;
     }
 
-    public String toString() {
+    @SuppressLint("DefaultLocale")
+	public String toString() {
```
bobzilladev, line 80: nit: Odd spacing

wiglewifiwardriving/src/main/java/net/wigle/m8b/geodesy/utm.java
```diff
@@ -179,7 +181,8 @@ public static double asinh(double x){
     }
     
 
-    public String toString() {
+    @SuppressLint("DefaultLocale")
+	public String toString() {
```
bobzilladev, line 185: nit: odd spacing

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/listener/BluetoothReceiver.java
```diff
@@ -930,7 +930,11 @@ public void onScanFailed(int errorCode) {
                         }
                     }
                     break;
-                default:
```
bobzilladev, line 933: Probably should still have a default case, maybe just to log so not silently failing?

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/util/SearchUtil.java
```diff
@@ -105,87 +105,80 @@ public static String setupQuery(final View view, final Context context, final bo
             }
 
             try {
-                switch (id) { //TODO: need to fix switch/case before Gradle 8
```
bobzilladev, line 108: Going from switch to if-else trees feels so lame. ohwell.

bobzilladev (approved):
A few comments. Glad to see so much cleanup! If you do more, can you do different PR's for different flavors of fixes, this was a daunting one 😄

wiglewifiwardriving/src/main/java/net/wigle/m8b/geodesy/mgrs.java
rksh: ah, external library from m8b had tabs instead of spaces.

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/listener/BluetoothReceiver.java
rksh: yeah, it warns, but I'll re-add it (basically the lint error was "you're not addressing all these cases specifically")

PR #659, opened 2023-12-20
[ENH] Massive linting

rksh:
the actual linting PR.

- fixed issues that would prevent us from migrating to gradle 8
- fixing layout overlaps
- got rid of some spurious warnings
- handling Locales for string generation
- wrapping build-specific constants with SDK version checks
- getting rid of no-longer-needed SDK version checks
- mass-moving-about of assets which were mis-located
- elimination of unused resources
- creation of monochrome launcher icon packs
- resizing of launcher icon to avoid bleed
- fixing some layout ID conflicts
- ellipsizing some raw ...s
- fixed leaks/unsafe calls in `MapUtils`


wiglewifiwardriving/src/main/java/net/wigle/m8b/geodesy/mgrs.java
```diff
@@ -75,8 +75,7 @@ public static mgrs fromUtm(utm u){
 	
 	return m;
     }
-
-    @SuppressLint("DefaultLocale")
+	@SuppressLint("DefaultLocale")
 	public String toString() {
```
bobzilladev, line 79: I don't really care, but this still is odd in git diffing, now both L78 and L79 line up with the return statement in L80 instead of the close-bracket on L81.

bobzilladev (approved):

PR #661, opened 2023-12-25
[ENH] Upload refresh in progress

rksh:
Allows the uploads screen to self-update on a 30s polling timer if left up, and if user stays on the "first page" of the upload results.

Also a small layout cleanup in map pin caption text.


bobzilladev (approved):
Refresh Your Thirst

PR #662, opened 2023-12-25
[FIX] Small layout fixes

rksh:
xmas linting - suppressing a couple of warnings and fixing some inefficient layout.


bobzilladev (approved):
hero to 0dp

PR #664, opened 2023-12-28
[ENH] Map fragment enhancements

rksh:
The network rendering behavior has been confusing for a long time.
This is one of a series of improvements to make network type, encryption status, and description more clear during clustering.
This adds:
- Pins colored according to network type
- Text Bubbles colored according to network type
- Icons indicating BT/BLE, 5G vs. other cell, and WiFi encryption status
- defaults to BSSID in the event of empty/null SSID in label.
- increase in labelled-nets from 1/15th to 1/10th of total net cache.

Very open to further refinement, this is just a first step.


bobzilladev (approved):
lgtm

PR #665, opened 2023-12-28
[ENH] Drawables: revenge of the lint

rksh:
A bunch of little visual linting checks. Grouped together because none of these reall bear much review.
- `xhdpi` drawable asset directory was missing a lot of versions
- an old, low-res grayscale planet icon was doing mixed/triple duty at points in the UI
- android 24-29 had a mismatched icon in notification bar
- DP sizes in icons were inconsistent across DPI folders
- removing a lint warning in our port of the map bubble UI (because we learned it from Google)
- an errant dimens entry


bobzilladev (approved):
i assume it looks good!

PR #666, opened 2024-01-01
[FIX] fix for recent lyout problems with masked text edit

rksh:
for some reason, this was absolutely breaking the masked text edit hint.
also threw in some related cleanup on warnings.


bobzilladev (approved):
If you see your pull request this weekend, be sure and tell her, SATAN 😈

PR #667, opened 2024-01-05
[ENH] icon complexity/speed hacks

rksh:
replacing 
- the inefficient vector planet with scaled images
- large VP yellow W vector with same vector - but smaller viewport.

also a few bits of linting that came up during the process.


bobzilladev (approved):
iconoclastic

PR #668, opened 2024-01-14
[[ENH] Topbar redesign

rksh:
user feedback: "Run" is confusing, and it's easy to misunderstand that networks are "lost" on app stop/crash.
solution:
retool topbar (but not dash) to highlight new nets since last run, omit "Run"


wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/ListFragment.java
```diff
@@ -140,6 +141,8 @@ public static class LameStatic {
     private boolean animating = false;
     private AnimatedVectorDrawableCompat scanningAnimation = null;
 
+    private SharedPreferences prefs;
```
bobzilladev, line 144: hrm, isn't it not a happy practice to keep a reference to prefs longer term, vs grabbing current every time?

bobzilladev: nm, guess it's all cool now man

uhtu (approved):

PR #672, opened 2024-01-20
[FIX] for GPX and M8B export thread queue and cache paths

rksh:
- at-most-one Thread queue was failing to respect job classes
- cache paths were missing a "/"
- added cancel-ability (not yet wired)

### DIFF
--- a/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/GpxManagementActivity.java
+++ b/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/GpxManagementActivity.java
@@ -177,8 +177,8 @@ public class GpxManagementActivity extends ScreenChildActivity implements PolyRo
             case EXPORT_GPX_DIALOG: {
                 if (!exportRouteGpxFile(exportRouteId)) {
                     Logging.warn("Failed to export gpx.");
-                    WiGLEToast.showOverFragment(this, R.string.error_general,
-                            getString(R.string.gpx_failed));
+                    //WiGLEToast.showOverFragment(this, R.string.error_general,
+                    //        getString(R.string.gpx_failed));
                 }
                 break;
             }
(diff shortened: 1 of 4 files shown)


wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/GpxManagementActivity.java
```diff
@@ -177,8 +177,8 @@ public void handleDialog(int dialogId) {
             case EXPORT_GPX_DIALOG: {
                 if (!exportRouteGpxFile(exportRouteId)) {
                     Logging.warn("Failed to export gpx.");
-                    WiGLEToast.showOverFragment(this, R.string.error_general,
-                            getString(R.string.gpx_failed));
+                    //WiGLEToast.showOverFragment(this, R.string.error_general,
+                    //        getString(R.string.gpx_failed));
```
bobzilladev, line 181: Should the user still be notified of this?

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/background/UniqueTaskExecutorService.java
```diff
@@ -21,25 +25,89 @@ public UniqueTaskExecutorService(int nThreads) {
...
 
     @Override
     protected void beforeExecute(Thread t, Runnable r) {
-        Logging.debug("===>EXECUTING: "+r.getClass());
-        current = r;
+        if (r instanceof WrappedTask) {
+            WrappedTask wt = (WrappedTask)r;
+            Logging.debug("===>EXECUTING: " + wt.getWrappedClass());
+            current = r;
+        }
```
bobzilladev, line 42: Should there be a log if this elses for some reason?

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/background/UniqueTaskExecutorService.java
```diff
@@ -21,25 +25,89 @@ public UniqueTaskExecutorService(int nThreads) {
...
     protected void afterExecute(Runnable r, Throwable t) {
-        jobClasses.remove(r.getClass());
-        current = null;
-        Logging.debug("<===EXECUTED: "+r.getClass());
+        if (r instanceof  WrappedTask) {
+            WrappedTask wt = (WrappedTask)r;
+            jobClasses.remove(wt.getWrappedClass());
+            current = null;
+            Logging.debug("<===EXECUTED: " + wt.getWrappedClass());
+        }
```
bobzilladev, line 52: Same else notion.

bobzilladev (approved):
Seems ok, uhtu has been the one digging in on this goo though.

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/GpxManagementActivity.java
rksh: the only case where we return false turns out to be the one we can (and are) more specific about inside the method (and the specific followed by the non-specific is annoying)

uhtu (approved):
lgtm (modulo bobz' else clause notes)

PR #675, opened 2024-01-21
[ENH] making missing start/end padding explicit

rksh:
just lint.

- largely making 0dp corresponding Start/Ends explicit.
0 useing the WiGLE service name for the icon in the toast, because I guess that's correct.


bobzilladev (approved):
android:reviewComments="0dp"

PR #676, opened 2024-01-21
[FIX] replacing leaking asyncTask

rksh:
Fixing an actual  issue (leaking static) instead of just linting for once.
Also removing deprecated menu-item layout cmds.

### DIFF
--- a/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/UserStatsFragment.java
+++ b/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/UserStatsFragment.java
@@ -8,14 +8,13 @@ import android.graphics.Bitmap;
 import android.graphics.BitmapFactory;
 import android.graphics.Color;
 import android.media.AudioManager;
-import android.os.AsyncTask;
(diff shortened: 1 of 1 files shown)


wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/UserStatsFragment.java
```diff
@@ -223,7 +223,9 @@ public void handleUserStats(final View view, final Resources resources, final St
             }
         }
         if (null != stats.getImageBadgeUrl() && !stats.getImageBadgeUrl().isEmpty()) {
-            new DownloadBadgeImageTask(view.findViewById(R.id.badgeImage)).execute(UrlConfig.WIGLE_BASE_URL+stats.getImageBadgeUrl());
+            ExecutorService executor = Executors.newSingleThreadExecutor();
```
rksh, line 226: this really is single use - this activity doesn't even have a "swipe to refresh" command.

uhtu (approved):

bobzilladev (approved):

PR #678, opened 2024-01-23
[ENH]  layout tweak

rksh:
Makes layout work better for languages where list_waiting_gps is a long string. Testing in french is a good reminder.


bobzilladev (approved):
balance on the center of gravity

PR #679, opened 2024-01-23
[FIX] RankStatsFragment ISE fix candidate

rksh:
here's a pass at making all these safe. the only reported exception is in RankStatsFragment which makes this feel a little bit like overkill, but there's reason to believe we're taking this risk in multiple places.

### DIFF
--- a/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/DataFragment.java
+++ b/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/DataFragment.java
@@ -39,6 +39,7 @@ import android.widget.Spinner;
 import android.widget.TextView;
 
 import static net.wigle.wigleandroid.MainActivity.ACTION_GPX_MGMT;
+import static net.wigle.wigleandroid.MainActivity.getMainActivity;
 import static net.wigle.wigleandroid.background.GpxExportRunnable.EXPORT_GPX_DIALOG;
 
 import com.google.android.material.textfield.TextInputLayout;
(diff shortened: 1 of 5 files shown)


wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/DataFragment.java
```diff
@@ -526,15 +530,15 @@ public void handleDialog(final int dialogId) {
                 if (!exportM8bFile()) {
                     Logging.warn("Failed to export m8b.");
                     WiGLEToast.showOverFragment(getActivity(), R.string.error_general,
-                            getString(R.string.m8b_failed));
+                            getActivity().getResources().getString(R.string.m8b_failed));
```
uhtu, line 533: is there a reason to be more confident about this getActivity not returning null (contra :466)?

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/RankStatsFragment.java
```diff
@@ -126,7 +126,10 @@ public void onAuthenticationRequired() {
                     public void onTaskCompleted() {
                         stopAnimation();
                         if (userDownloadFailed) {
-                            WiGLEToast.showOverFragment(a, R.string.upload_failed, getString(R.string.dl_failed));
+                            if (null != a) {
```
uhtu, line 129: i think a was already nullchecked at :113 ?

uhtu (approved):

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/RankStatsFragment.java
rksh: quite right - I was being cautious because it's in a callback, but the reference is final, it'll be fine.

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/DataFragment.java
rksh: not a good one.

bobzilladev (approved):

uhtu (commented):
👍

PR #681, opened 2024-01-24
[FIX] ProgressPanelRunnable safety

rksh:
Nothing inside this block makes a difference if the progress panel is null.

### DIFF
--- a/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/background/ProgressPanelRunnable.java
+++ b/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/background/ProgressPanelRunnable.java
@@ -74,25 +74,27 @@ public abstract class ProgressPanelRunnable implements AlertSettable {
             @Override
             public void run() {
                 //DEBUG: Logging.info("progress: "+percent);
-                if (100 == percent) {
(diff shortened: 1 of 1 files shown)


wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/background/ProgressPanelRunnable.java
```diff
@@ -74,25 +74,27 @@ protected void onProgressUpdate(Integer percent) {
...
+                    }
+                    if (null != executorService) {
+                        final int curSize = executorService.getQueue().size();
+                        if (curSize != lastTaskQueueDepth) {
+                            if (curSize == 0) {
+                                pp.hideQueue();
+                            } else {
+                                pp.setQueue(activity.getString(R.string.queued_jobs, curSize));
+                            }
+                            lastTaskQueueDepth = executorService.getQueue().size();
```
uhtu, line 96: this mightshould be ```lastTaskQueueDepth = curSize``` to avoid racing the check/use?

rksh: common sense: i was last in line when they were handing it out.

uhtu (approved):

bobzilladev (approved):

PR #685, opened 2024-01-29
[FIX] fixed-length -> dynamic array list for service UUIDs

rksh:
fixing a production crash by allowing the list to grow after loading


bobzilladev (approved):
more new's is good news

Issue #683, opened 2024-01-28
Send broadtcast on sucessful upload


grim0us:
It would be great to know when an upload has finished successfully. Would it be possible to create a broadcast to notify other apps, e.g. Tasker, when a successful upload has been completed?

[Tasker intents](https://tasker.joaoapps.com/userguide/en/intents.html) under "Receiving Intents"

rksh:
will be included in https://github.com/wiglenet/wigle-wifi-wardriving/releases/tag/2.86 and up

PR #686, opened 2024-02-06
[ENH] Upload complete broadcast

rksh:
Implementing #683 - sending a broadcast intent on upload success or failure.
These are (respectively) `net.wigle.wigleandroid.UPLOAD_COMPLETE` and `net.wigle.wigleandroid.UPLOAD_FAILED`.

As a part of this change, I've also moved the `OkFileUploader` core call contents to WiGLEApiManager, our API implementation class. This necessitated numerous changes and begged certain cleanups, including:
- making the global state api manager instance public, since we reference it outside of its package
- cleanups of some literal strings and statuses, including two non-enumerated/conflicting integer states
-  eliminating some unused, duplicative code
- a few safety checks that really should have been there from the start
- parsing and inclusion of the upload results, including transIds
- state handling cleanup - this is still not perfect, but should be cleaner

Some of the changes here may benefit from further refactoring, but this change set is getting large, and further changes (e.g. combining our various progress interaction implementations) would increase risk.


bobzilladev (approved):
lgtm, except for one un-intent-ional thing to fix.

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/background/ObservationUploader.java
```diff
@@ -238,69 +248,90 @@ private Status doUpload( final Bundle bundle )
...
+                            editor.putLong( PreferenceKeys.PREF_DB_MARKER, maxId );
+                            editor.putLong( PreferenceKeys.PREF_MAX_DB, maxId );
+                            editor.putLong( PreferenceKeys.PREF_NETS_UPLOADED, dbHelper.getNetworkCount() );
+                            editor.apply();
+                            final UploadReseponse.UploadResultsResponse uploadResults = response.getResults();
+                            List<String> transIds = uploadResults.getTransids().stream()
+                                    .map(UploadReseponse.UploadTransaction::getTransId)
+                                    .collect(Collectors.toList());
+                            if (transIds.size() > 0) {
+                                intent.putExtra("transIds", transIds.toString());
```
uhtu, line 274: do we care about doing this toString() more than once?
i suppose these are not frequent.

uhtu (approved):

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/background/ObservationUploader.java
rksh: it's sillier than that- the client doesn't currently support multiple CSVs/archive upload - so transIds will always be a single-item list. Just doing this to correctly process the API response should we start supporting multi-archive. Cleaning up.

PR #688, opened 2024-02-13
[FIX] 5g icon fixes - size and visibility

rksh:
finally found somewhere I could test NR:
- increased icon size and legibility
- corrected some cases


wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/ui/NetworkIconGenerator.java
```diff
@@ -247,7 +248,6 @@ private int getIconId(NetworkType type, int crypto) {
                     case Network.CRYPTO_WPA3:
                         return R.drawable.wpa3_ico;
                 }
-                return R.drawable.cell_5g;
```
bobzilladev, line 250: Interesting it will let you remove a default here, compiler must check that switch is exhaustive?

bobzilladev (approved):

uhtu (approved):

PR #691, opened 2024-02-20
[FIX] npe checks on pp handling

rksh:
just safety checks.


bobzilladev (approved):
i'm down with null pp

PR #692, opened 2024-03-08
[ENH] SVGs look much better.

rksh:


bobzilladev (approved):
we _do_ need those svgy badges!

PR #697, opened 2024-04-01
[FIX] Null check route list

rksh:
these are almost certainly race conditions/corner cases on shutdown, but might as well check.


bobzilladev (approved):
null checks are part of this balanced breakfast

PR #705, opened 2024-06-30
fix npe in CellSearchResponse

bobzilladev:
java.lang.NullPointerException: Attempt to invoke virtual method 'java.lang.String java.lang.String.toUpperCase(java.util.Locale)' on a null object reference
at net.wigle.wigleandroid.model.api.CellSearchResponse.asNetwork(CellSearchResponse.java:100)


rksh (approved):
+1

Issue #703, opened 2024-06-18
Bug and feature simultaneously


suchencjusz:
Possibility for setting custom user agents

![Screenshot_20240618-210407_WiGLE Wifi Wardriving](https://github.com/wiglenet/wigle-wifi-wardriving/assets/34921955/ddd2ac6e-6d53-4f57-ac9b-56b2db8d0d09)

rksh:
We will fix that, but probably not make it customizable - it's useful to us in understanding site traffic (the only thing the client hits with the OKHttp client is WiGLE's servers). Why would you want a custom user agent?

suchencjusz:
I worded it wrong, I don't really need the functionality of changing the user agent.  but it seems to me that information about the system kernel is not necessary

slammingprogramming:
@rksh with regard to issue #730 I would like to find out if this issue of custom user agents being out-of-scope due to the current state of the app not reaching servers other than wigles would still stand if we implement the proposed solution for the aforementioned issue referenced.

rksh:
The purpose of many of the open issues is to solicit public interest. It's quite unusual for a new contributor to voluteer a major code contribution as well as comment/offer to take on the majority of open tickets in a project (even those that are not currently slate/intended for inclusion). Are you a bot or intending to execute all these changes using a bot? Please note our contribution policy: https://github.com/wiglenet/wigle-wifi-wardriving/blob/main/CONTRIBUTING.md

slammingprogramming:
Just notating here, since I already replied to that question on issue #730 that I am a human and all that, check my reply there for more elaboration. Short version is: I'm a human who uses WiGLE and I just intend to contribute to these since each I replied to I personally find useful.

PR #706, opened 2024-07-23
[FIX] filtering non-safe ASCII header characters from user agent, centralizing

rksh:
fix for https://github.com/wiglenet/wigle-wifi-wardriving/issues/703

filters non-ascii header chars if present in the elements we use to construct the user agent header and DRYs UserAgent generation into the WiGLEApiManager class.


bobzilladev (approved):

uhtu:
👍

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/net/WiGLEApiManager.java
```diff
@@ -88,6 +88,7 @@ public class WiGLEApiManager {
                     System.getProperty("os.name") + " " +
                     System.getProperty("os.version") +
                     " [" + System.getProperty("os.arch") + "]";
+            javaVersion = javaVersion.replaceAll("[^\\x00-\\x7F]", "");
```
uhtu, line 91: xkcd.com/1171/

PR #721, opened 2024-12-15
[ENH] more icon fixes

rksh:
Icons have been confusing for a long time.

- switching wrench to filter on ListFragment (since it's the list filtering tool)
- changing map layer from map icon on map fragment


bobzilladev (approved):
iconic

PR #725, opened 2025-01-15
[ENH] line width reductions at bobz's request

rksh:
2->1.5 width for some key icons


bobzilladev (approved):
I do personally like the look of that a lot more 🌶️

PR #737, opened 2025-07-19
[FIX] fix for menu weirdness

rksh:
what a strange bug. not refactoring to use MenuHost at this point, but we need to fix that too.
based on a report from NanoTrasen on the forums: https://wigle.net/phpbb/viewtopic.php?t=3363.


wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/MappingFragment.java
```diff
@@ -1015,6 +1002,7 @@ public boolean onOptionsItemSelected(@NonNull final MenuItem item ) {
                         edit.apply();
                         googleMap.setMapType(newMapType);
                     });
+                    return true;
```
rksh, line 1005: the actual fix.

bobzilladev (approved):

bobzilladev (commented):
HA! I said I thought that felt like a switch/default issue

PR #742, opened 2025-07-25
[ENH] show the vendor for the OUI prefix for addres filter elements

rksh:
Filters are already hard enough to use - why not give people a hint that they got the right OUI?


bobzilladev (approved):
❤️

PR #755, opened 2025-09-01
[FIX] fixes failure to update-  but is this the right way?

rksh:
@uhtu points out the list doesn't refresh on add of MAC/OUI without exiting
this is sort of a brute-force update. is there something more elegant?


bobzilladev (approved):
Seems fine, only on human input so rare, slow and correct ftw.

PR #760, opened 2025-10-07
[ENH] no more bar, but we miss the left arrow

rksh:
we all miss the left-arrow from the now-removed action bar on the network detail screen. shoe-horning it back in next to the SSID and network icon.

Also, adding an option to copy SSID and BSSID with long-press, because maybe those are useful (from user request).


bobzilladev (approved):
left arrow and clippy ftw.

Issue #741, opened 2025-07-24
Please add a filter toggle for Wi-Fi


lukejenkins:
Currently to limit the display to  only cell and/or Bluetooth, a user has to uncheck all of the Wi-Fi auth types. Having a master "Wi-Fi" toggle would be a nice improvement.

rksh:
Addressed in #773

PR #773, opened 2025-11-07
[ENH] All wifi filter mode

rksh:
Adds a single "`show wifi`" control that rolls up the sub-checkboxes.
- when checked, automatically checks any unchecked sub-items in `open`, `wep` , and `wpa`
- when when unchecked, automatically unchecks and checked sub-items in `open`, `wep` , and `wpa`
- when any of the sub-items is checked individually, the state of the `show wifi` box be updated by the implied result

also adding a small logging improvement in the service of diagnosing foreground service problems in Android SDK 34+

Attempted using `MaterialCheckBox` - but that necessitated updating theme basis to `Theme.Material3...` which breaks a number of app styles.

It's remarkable how many changes such a simple feature takes!

addresses https://github.com/wiglenet/wigle-wifi-wardriving/issues/741

### DIFF
--- a/wiglewifiwardriving/src/main/res/values-ar/strings.xml
+++ b/wiglewifiwardriving/src/main/res/values-ar/strings.xml
@@ -468,4 +468,5 @@
     <string name="ble_appearance_title">مظهر</string>
     <string name="alert_filter_button">تنبيه بشأن أجهزة MAC/OUIs…</string>
     <string name="status_too_many">كثير جدًا في الإطار الزمني</string>
+    <string name="filter_show_wifi">إظهار WiFi</string>
 </resources>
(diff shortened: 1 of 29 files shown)


wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/MappingFragment.java
```diff
@@ -1179,4 +1191,17 @@ public static int getRouteColorForMapType(final int mapType, final boolean night
         }
         return OVERLAY_DARK;
     }
+
+    private static void updateWifiGroupCheckbox(final View view) {
```
uhtu: DRY wants these to be pulled into something, but i think that's a future-problem.

uhtu (approved):

PR #777, opened 2025-11-12
[ENH] Wifi network view tools

rksh:
rethinking the tools for MAC and OUI filtering, blocking, and alerting.
provides the bottom bar in both BT and WiFi

### DIFF
--- a/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/NetworkActivity.java
+++ b/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/NetworkActivity.java
@@ -1315,24 +1315,24 @@ public class NetworkActivity extends ScreenChildActivity implements DialogListen
         return results.toString();
     }
 
-    private static void checkChangeHandler(final boolean checked, final String ssid, final boolean ouiMode,
+    private static void checkChangeHandler(final boolean checked, final String bssid, final boolean ouiMode,
                                             final List<String> currentAddresses, String prefKey, SharedPreferences prefs) {
-        if (ssid != null) {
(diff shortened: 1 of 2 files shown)


wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/NetworkActivity.java
```diff
@@ -1290,4 +1314,25 @@ private static String characteristicDisplayString(final Map<String, String> char
         }
         return results.toString();
     }
+
+    private static void checkChangeHandler(final boolean checked, final String ssid, final boolean ouiMode,
+                                            final List<String> currentAddresses, String prefKey, SharedPreferences prefs) {
```
bobzilladev: Looks like all callers are talking about `bssid`, so would probably change all the `ssid` variable names in here.

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/MacFilterActivity.java
```diff
@@ -196,11 +196,22 @@ public static boolean addEntry(List<String> entries, SharedPreferences prefs,
...
             return true;
         }
         return false;
     }
+
+    public static void updateEntries(List<String> entries, SharedPreferences prefs, final String filterKey) {
+        Gson gson = new Gson();
+        String serialized = gson.toJson(entries.toArray());
+        final SharedPreferences.Editor editor = prefs.edit();
+        editor.putString(filterKey,serialized);
```
bobzilladev: Was concerned about a limit to how much you can put in here, but looks like it just goes into an xml document that's loaded on startup, so limit is heap space, so load it up!

PR #778, opened 2025-11-14
[ENH] Passpoint logo display

rksh:
a take on doing https://github.com/wiglenet/wigle-wifi-wardriving/pull/648 using an overlay badge rather than a fresh icon.
Seems like these changes might be a little heavy...


rksh:
with apologies for my homebrew iconography - as far as I can tell, there's no formal HS2.0 / passpoint logo?

PR #781, opened 2025-11-27
[ENH] ANR minimization

rksh:
this old deprecated call is 1. redundant and 2. the source of our most common ANR. let's keep it confined to pre-deprecation devices only and minimize the pain.


bobzilladev (approved):
lock it away

PR #785, opened 2025-12-24
[FIX] adding desugaring

rksh:
Keep old android devices working, prevent ClassNotFounds on JDK 1.8+ calls


bobzilladev (approved):
F.U. SUGAR

PR #790, opened 2026-01-08
[FIX] because new isn't always better. 

rksh:
with info from[]( https://stackoverflow.com/questions/79128940/mysterious-accessibility-related-crash) was able to install Vigilante on a test device and reproduce the issue. Isolated to versions 1.12.0 - 1.14.0-alpha8 - left a note in the gradle file warning against upgrading without testing.


rksh:
NB: will require some testing to make sure this doesn't break the edge-to-edge changes that originally inspired the upgrade.

PR #796, opened 2026-02-06
[ENH/FIX] Record cell info anr

rksh:
at long last, refactored cell out of wifi receiver.
this is a wonky/large change, but we were having ANR problems as a result of the binder calls we've now pushed across the executor boundary.


bobzilladev (approved):
Seems cromulent, great to get that all in separate code on its own timer thread. Surprised `tele.getAllCellInfo();` is that expensive though.

PR #798, opened 2026-02-11
[FIX] background-thread for play console-reported ANR.

rksh:
Play console detected an ANR when clearing previous route on start-up in "show current" and "don't log routes" case.
Moves clear operation to a background thread.


bobzilladev (approved):
Thread them all!

Issue #802, opened 2026-03-04
Feature Request - Ability to silence OUI/MFGR alerts in a toggle-able way.


Azzybana:
I'm on 2.104, Android, installed via Obtanium, using latest release.

In Network List, under Filters, there was added Alert on MAC/OUI and one for BLE MFGR ID.

I've input about 20 OUIs, and 6 MFGR IDs. 

I hear a failed thump all the time, or success ping near a device.

I also use my phone as my AAC as I'm mute ie non verbal. Because my phone talks for me, I frequently have to max the volume. 

There's no way to silence the thump of a fail, or the ping of an alert, except to delete all of them, that I've found. 

Suggestion for a solution...

Either have it respect the "Filters enabled" and only make sounds when that's on, or to add a setting under Settings which I'd envision as a tri-state drop down of "Alert Sound" or whatever, with None, Device Found, All. (None muting alerts, Device Found only playing the success sound.)

If I missed the setting to toggle this, I apologize. It's not in the manual that it even exists yet. 

Also the sound continues playing, always a fail sound (even when a nearby device is present), when scanning is disabled. It does change tempo with movement if movement is set differently than stationary. 

It shouldn't be playing any sounds when it's not even looking.

My current workaround is simply close the app.

PR #803, opened 2026-03-07
feat: making BSSID/MAC/MfgrId sounds obey the global mute.

rksh:
addressing requests in issue #802


bobzilladev:
I forgot too, was poking around at who checked that pref and was like..oh!

bobzilladev (approved):
Approving to unblock, one question though.

PR #819, opened 2026-04-09
Remove certificatetransparency-android before goog breaks the app

bobzilladev:
Joy: https://googlechrome.github.io/CertificateTransparency/apps_3p_libraries.html

This compiles, now to test.
Xml from: https://developer.android.com/privacy-and-security/security-config#certificateTransparency


bobzilladev:
Closing to consolidate

PR #820, opened 2026-04-10
chore: google is breaking 3p CERTTRANS libraries in favor of their own

rksh:
maybe it's safer? maybe it's lock-in?
regardless, we can't release without removing the old library.

### DIFF
--- a/wiglewifiwardriving/build.gradle
+++ b/wiglewifiwardriving/build.gradle
@@ -108,7 +108,6 @@ dependencies {
     implementation 'com.squareup.okhttp3:okhttp:4.11.0'
     implementation 'com.github.edwardstock:edittext-mask:1.0.7'
     implementation 'com.goebl:simplify:1.0.0'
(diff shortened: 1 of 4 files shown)


bobzilladev:
lol, didn't see my chat i guess: https://github.com/wiglenet/wigle-wifi-wardriving/pull/819

bobzilladev:
I turned it on for everything, we pry don't need to exclude to just our domain. For instance if we allow other configurations in the future.

rksh:
my concern was the mapLibre tile fetches - but that exception I made will only matter in SDK 36 - we know we conform to CERTTRANS already. (SDK 37 will require it for everything regardless)

bobzilladev:
I guess we can always turn it all on later. Is the trust-anchor system just the default, if so do we really need it?

bobzilladev (approved):
One outstanding comment, approving to unblock.

PR #823, opened 2026-04-22
fix: technically, you could exceed -20 and screw up the playback pitch

rksh:
no attenuation? no problem!


bobzilladev (approved):
dooby dooby doo

PR #824, opened 2026-04-25
fix: textual fallback for indeterminate progress when system animatio…

### LINKED ISSUE #699: Confusing loading state in Battery Saver
Dear developer,

I noticed that when I enable Battery Saver on Android 8.1, the indeterminate `ProgressBar`s in the app are not properly shown. This is a known problem in Android API level <28, see e.g. [this StackOverflow question](https://stackoverflow.com/questions/35221706/progressbar-disappears-in-battery-saver-mode-android-5-x). Battery Saver disables animations, also on progress bars on these versions. This is quite confusing because the loading state is not properly represented. It is fixed in later Android versions, where progress bars do appear and animate.

Looking at the code, indeterminate progress bars are created here:

```
className='net.wigle.wigleandroid.background.AbstractBackgroundTask', lineNumber=169
className='net.wigle.wigleandroid.ProgressPanel', lineNumber=30
```

To fix this issue, one can check [ValueAnimator.areAnimatorsEnabled()](https://developer.android.com/reference/android/animation/ValueAnimator) and provide a different UI element, such as a text label, when animations are disabled.

BennyWaitWhat:
Closes #699.

On API 21-27 with `animator_duration_scale = 0` (Battery Saver, Reduced Animations), `ProgressBar.setIndeterminate(true)` renders nothing. The reporter's diagnosis in #699 matches the current code in `ProgressPanel` and `AbstractBackgroundTask`.

Detect via `ValueAnimator.areAnimatorsEnabled()` (API 26+). When animators are off, hide the bar and cycle trailing dots on the status label (`Working`, `Working.`, `Working..`, `Working...`) with a `Handler.postDelayed` loop. `setProgress(int)` and `hide()` tear it down. Centralized in `ProgressPanel` so both `AbstractBackgroundTask` and `ProgressPanelRunnable` get it. Same patch removes two redundant direct `progressBar.setIndeterminate(true)` calls in those callers; they duplicated the `pp.setIndeterminate()` immediately following.

Why a `Handler` not another `ValueAnimator`: `areAnimatorsEnabled()` is the detection signal, and the system has disabled `ValueAnimator`, which is what the fallback would otherwise rely on. `Handler.postDelayed` is the standard escape hatch.

API 24/25 keep the original behavior. The underlying bug is rare there and there's no public way to detect animator scale = 0 without reflection. Easy to extend if reports come in.

FOSS portability: `ProgressPanel` has no `Foss*` sibling (shared between `main` and `foss-main`). All three changed files use only `android.*` APIs. Mechanical port.

Manual test plan:
1. Animations off. Developer Options, set "Animator duration scale" to "Animation off". Trigger a background task. Expect bar hidden, label cycling dots ~every 0.4s.
2. Animations on. Scale to "1x". Expect normal animated bar, static label.
3. Determinate progress. Long upload. Expect bar visible and progressing, no dots.
4. Battery Saver. Enable in Settings on a device that flips animator scale. Same as #1.

Files: `ProgressPanel.java` (fallback), `AbstractBackgroundTask.java` and `ProgressPanelRunnable.java` (redundant call removed). No XML, strings, or manifest changes.

### DIFF
--- a/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/ProgressPanel.java
+++ b/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/ProgressPanel.java
@@ -1,5 +1,9 @@
 package net.wigle.wigleandroid;
 
+import android.animation.ValueAnimator;
+import android.os.Build;
+import android.os.Handler;
+import android.os.Looper;
 import android.os.Parcelable;
 import android.view.View;
 import android.widget.LinearLayout;
@@ -19,6 +23,15 @@ public class ProgressPanel {
     private final ProgressBar progressBar;
     private final TextView queueLabel;
 
+    /** Base message text without any animated dot suffix. */
+    private String baseMessage = "";
+    /** Lazily created; only used when system animations are disabled. */
+    private Handler dotAnimationHandler;
+    private Runnable dotAnimationRunnable;
+    private int dotCount;
+
+    private static final long DOT_ANIMATION_INTERVAL_MS = 400L;
+
     public ProgressPanel(LinearLayout layout, TextView progressLabel, ProgressBar progressBar, TextView queueLabel) {
         this.container = layout;
         this.progressLabel = progressLabel;
@@ -31,16 +44,26 @@ public class ProgressPanel {
     }
 
     public void hide() {
+        stopDotAnimation();
         setProgress(0);
         progressBar.setIndeterminate(true);
         container.setVisibility(View.GONE);
     }
 
     public void setMessage(final String text) {
-        if (null != progressLabel) progressLabel.setText(text);
+        baseMessage = (text == null) ? "" : text;
+        if (null != progressLabel) {
+            progressLabel.setText(buildLabelText());
+        }
     }
 
     public void setProgress(final int progress) {
+        // Determinate progress works fine even when system animations are off;
+        // tear down any textual fallback and ensure the bar is visible.
+        stopDotAnimation();
+        if (null != progressBar) {
+            progressBar.setVisibility(View.VISIBLE);
+        }
         progressBar.setIndeterminate(false);
         progressBar.setProgress(progress);
     }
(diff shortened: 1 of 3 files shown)


BennyWaitWhat:
Quick on-device sanity check on a Pixel 9 Pro (Android 14, API 34) with `animator_duration_scale = 0` system-wide before launch:

- Patched debug APK installs and launches cleanly. No crashes, no regressions on the data tab.
- Triggered an upload. The inline progress panel appears and the determinate path renders correctly. Caught it mid-task at "Writing File 100.0%" with the orange bar visible, and "Downloading 22.7%" during a stats sync.

What I deliberately did not capture: the indeterminate-phase frame itself (bar hidden, dots cycling). The window where `setIndeterminate()` is active before `setProgress(int)` overrides it is sub-100ms in this flow, too short for `screencap` even with chained shell commands. Worth flagging that this device is API 34, where the original `ProgressBar` rendering bug doesn't reproduce; the fallback path runs any time animations are off on API 26+, but a meaningful visual demo really wants an API 21-27 device or emulator. Happy to spin one up if useful before merge.

Determinate path: verified. Indeterminate path: code-reviewed, not yet visually demonstrated.

BennyWaitWhat:
Circling back on "indeterminate path: code-reviewed but not visually demonstrated yet". Got the frame this time.

Pixel 9 Pro, Android 16 (API 36), `animator_duration_scale = 0` system-wide before launch. Triggered Database -> Backup database, recorded at 10 fps.

**Indeterminate, fallback engaged** (~100 ms after the confirm dialog dismisses). Label reads `Backing up database...`, `ProgressBar` is `View.GONE`.

indeterminate, bar hidden

**Determinate, bar restored** (~100 ms later). Same label, bar back as a thin determinate fill. First `setProgress(int)` arrived, dots torn down.

determinate, bar visible

Dot cycle didn't advance: it's 400 ms per increment and `BackupRunnable` reports its first percentage well under that, so the captured indeterminate frame is at `dotCount = 0` (base text only). On API 21-27 where the `ProgressBar` bug actually repros, the same window would run through the dot cycle.

Also pushed a small amend: a one-line early-return in `stopDotAnimation()` so it no-ops when nothing is running. `setProgress(int)` calls it on every percent update; without the guard the helper rewrites the label `TextView` redundantly on the determinate hot path. Branch stays a single commit.

Determinate path: verified. Indeterminate fallback: now visually demonstrated.

bobzilladev:
Vote close this one, based on being unnecessary LLM noise. It did inspire us to make a contributing file though.

PR #826, opened 2026-04-25
chore: fix StartWigleAtBootReciever class name typo and "wierd" log

BennyWaitWhat:
Two small internal cleanups in one commit. No behavior change, no UI strings, no public API.

Class rename: `StartWigleAtBootReciever` to `StartWigleAtBootReceiver`. The `BroadcastReceiver` subclass in `net.wigle.wigleandroid.listener` has had a transposed `ie`/`ei` since inception. The `<receiver android:label="...">` attribute on `AndroidManifest.xml` line 361 is already spelled correctly, so the typo was caught in the user-facing label but missed on the class, its `.java` file, and the `android:name=".listener.StartWigleAtBootReciever"` reference. This PR brings them into agreement and matches Android's standard `*Receiver` convention.

Changes:
- `StartWigleAtBootReciever.java` to `StartWigleAtBootReceiver.java` (file rename via `git mv`, 93% history similarity).
- Class declaration updated inside the file.
- `AndroidManifest.xml` line 360: `android:name` updated.

No imports reference this class by name (only the manifest invokes it via the framework), so the rename is fully contained.

Log message: `"wierd ex:"` to `"weird ex:"` in `ListFragment.java` line 802, inside the `IncompatibleClassChangeError` catch handler. The comment on the line above ("yeah, saw this in the wild, who knows.") suggests this was a debug-session leftover. Logcat-only, never user-facing, but easy to fix while we're here.

### DIFF
--- a/wiglewifiwardriving/src/main/AndroidManifest.xml
+++ b/wiglewifiwardriving/src/main/AndroidManifest.xml
@@ -357,7 +357,7 @@
             </intent-filter>
         </receiver>
         <receiver
-            android:name=".listener.StartWigleAtBootReciever"
+            android:name=".listener.StartWigleAtBootReceiver"
             android:label="StartWigleAtBootReceiver"
             android:exported="true"
             android:permission="android.permission.RECEIVE_BOOT_COMPLETED">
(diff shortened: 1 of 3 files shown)


bobzilladev (approved):
Accepting, just speling fixes.

Issue #783, opened 2025-12-06
Unable to upload once outstanding networks exceed server-side upload size limit


Firecul:
I have been trying to upload for quite a while but I haven't been able to successfully. It's really getting to the point it'll be a problem. The writing file seems to get stuck at 36.9% in the ui but seemingly continues because it will eventually start to upload but fails whether it's on wifi or data.  
It does give me a file name but that file isn't available to me anywhere I can find as a non-rooted phone.  

Is there a way to save this file manually I haven't found so I can upload via a pc?

Pixel 6 Pro, v2.99

| Writing | Uploading |
| ----- | ----- |
| <img width="1440" height="3120" alt="Image" src="https://github.com/user-attachments/assets/122a8f82-32c4-4b91-8816-f1c2451f7c82" /> | <img width="1440" height="3120" alt="Image" src="https://github.com/user-attachments/assets/c3cca474-8a69-48af-a296-d02d640df426" /> |

(Yes I realise this is more like a double bug report 36.9% ui+upload fail but they might be related so decided to just open one.)

Frigyes06:
had this happen, logged out and now I can't log back in. I'm getting error 999

rksh:
so, just back of the envelope, an outstanding (un-uploaded) list of networks > 800k starts to get risky (if you have lots of BTLE with metadata and WiFi with RCOIs which richer CSV lines). By the time you reach 1.5m, it's likely, and at 1.9m it's a virtual certainty.

Firecul:
That doesn't sound like unreasonable numbers and is probably about in line with what I experienced.

rksh:
https://github.com/wiglenet/wigle-wifi-wardriving/pull/822 does not fix this issue directly, but _will_ help diagnose (and sets the stage to fix in a subsequent PR)

rksh:
fixed by auto-partitioning uploads based on estimated size. expect in release 2.109+

PR #831, opened 2026-04-30
fix: partitioning uploads to remedy #783.

rksh:
Breaks too-large uploads into multiple upload chunks. Also revises warning text in DB activity to warn about chunking instead of failure.

- Uses our previously-established max-file-size estimation values 
- actual upload sizes in practical testing with cell/bt/wifi all enabled: 170,721,334 - 154,603,855 bytes

There's a case to be made for some follow-on features: 
- changing new/un-uploaded number on the main screen (and in the DB) when we reset DB marker to zero (deemed outside the scope of this PR, but it's how I tested it)
- warning people who reset their DB marker because ew.
- refining upload size estimates
- CLEAN-UP: when you reset your DB marker and make N near-maximum size uploads to test this, it's WASTEFUL w/ local disk space.
- somehow I was credited by the server with 7 new bluetooth + location from re-uploading 10,125,009 wifi+cell+bt nets; suggests an upload race condition on BT CSV gen.
- do we ever need to check compressed artifact size (back of the envelope we're always safe, but could we hypothetically create a high-entropy enough file that GZ can't save us?

### DIFF
--- a/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/background/ObservationUploader.java
+++ b/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/background/ObservationUploader.java
@@ -11,7 +11,6 @@ import android.content.pm.PackageInfo;
 import android.content.pm.PackageManager;
 import android.database.Cursor;
 import android.os.Bundle;
-import android.preference.PreferenceManager;
 import androidx.fragment.app.Fragment;
 import androidx.fragment.app.FragmentActivity;
 
@@ -48,9 +47,9 @@ import java.nio.BufferOverflowException;
 import java.nio.ByteBuffer;
 import java.nio.CharBuffer;
 import java.nio.channels.ClosedByInterruptException;
-import java.nio.charset.Charset;
 import java.nio.charset.CharsetEncoder;
 import java.nio.charset.CodingErrorAction;
+import java.nio.charset.StandardCharsets;
 import java.text.DecimalFormat;
 import java.text.FieldPosition;
 import java.text.NumberFormat;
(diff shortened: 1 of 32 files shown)


wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/background/ObservationUploader.java
```diff
@@ -266,7 +268,9 @@ public void onTaskSucceeded(UploadReseponse response) {
                             status = Status.SUCCESS;
                             final SharedPreferences.Editor editor = prefs.edit();
                             editor.putLong( PreferenceKeys.PREF_DB_MARKER, maxId );
-                            editor.putLong( PreferenceKeys.PREF_MAX_DB, maxId );
+                            if (!countStats.rowsCapped) {
+                                editor.putLong( PreferenceKeys.PREF_MAX_DB, maxId );
```
bobzilladev, line 272: Seems like we should set it to the new value at least, so if anything dies while next file(s) should be going we don't lose that progress and end up uploading all of them again?

bobzilladev: Hrm, guess this is visual only, the marker is set. But maybe this should be updated if the current value is less than the new maxid?

bobzilladev (approved):
One comment

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/background/ObservationUploader.java
rksh: like:
```
if ( !countStats.rowsCapped && [currentValue] < maxId ) {
```
?

bobzilladev: Was thinking `if ( [currentValue] < maxId ) {`. If the number is higher, we can show it, regardless of capping? But maybe I'm a low-information voter here.

Issue #726, opened 2025-02-25
Bluetooth feature request: ability to differentiate BLE public vs. random addresses


XenoKovah:
I may just be missing it, but AFAICT currently the raw WiGLE data (as obtained by looking at the SQLite schema) only encodes whether a Bluetooth address is Classic/BR/EDR ("B") or BLE ("E"). However, BLE has both "public" addresses (which behave the same as classic, and the top 24 bits are an OUI), and "random" addresses (which have 3 sub-types, "random static", "random private resolvable", and "random private non-resolvable", which are differentiated based on their 2 MSBits).

The request is to introduce a new mnemonic (e.g. "R" or something like that if that's already used), to make it clear which BDADDRs are BLE random type (and then anyone wanting to know which specific type can just check the top 2 bits themselves. Or perhaps this could go in the capabilities field?)

XenoKovah:
> To date, we haven't discovered a way to get the Txadd/Rxadd bits from the PDU header via the Android LE ScanRecord. If my reading of the spec is right, we need these in addition to the address to conclusively determine public/random - is this correct?

Nominally. But it sounds like Android is exposing it some other way. I don't know anything about the Android BT stack, but if you're operating on ScanRecords, then these two entries seem relevant:

https://developer.android.com/reference/android/bluetooth/le/ScanRecord#DATA_TYPE_PUBLIC_TARGET_ADDRESS
https://developer.android.com/reference/android/bluetooth/le/ScanRecord#DATA_TYPE_RANDOM_TARGET_ADDRESS

Can those be tested against somehow on the discovered BDADDRs coming out of BLE scan results?

(And again, I don't really know Android, but that "added in API 33" link seems to imply that they're only available in Android 13 and higher? I'm currently running WiGLE on an Android 12 Pixel 3 which AFAIK doesn't support Android 13, so I *am* sensitive to not wanting to exclude large swaths of Android devices.)

XenoKovah:
Actually now that I think of it those names are also used for some rare advertisement types...

![Image](https://github.com/user-attachments/assets/9ec2abaa-1067-4886-b81a-cbb9790118a5)

So if those data types correspond to those spec things, then yes, I'm not really sure where one would find the public/random differentiator or whether it's exposed on Android.

XenoKovah:
Do you have access to BluetoothDevice objects described here? https://developer.android.com/reference/android/bluetooth/BluetoothDevice

[...]

PR #832, opened 2026-05-02
Revisiting the fix for issue #726 (distinguishing BLE Public vs. Random addresses.)

XenoKovah:
This is currently in a 'works for me' state for differentiation of BLE Public vs. Random addresses. I had Claude code this up, and also use Sniffle as a source of truth to verify the correctness of 60 seconds worth of data. Also updated the OUIs because it was missing an Espressif one, which was causing incorrect identification for the fallback logic due to lack of match on OUI for that.

Again, I didn't familiarize myself with the codebase, I just got it working on my own copy with Claude's help, so there might be better places to put things. And I don't know how much of a PITA it would be to change out the "E" for BLE to "P" for BLE Public addresses, just so the abbreviations made nice sense would be, so that's very much just a nice-to-have.

But I did see in the Claude summaries and can confirm some of the previous logic being wrong for distinguishing BLE Random sub-types based on the 3 MSBits (because it was using the 3 LSBits of the MSByte).


rksh:
the consequences of adding a new NetworkType for `BLE_RANDOM` extend beyond just the app; please resubmit a much narrower fix that doesn't dictate server-side changes / re-plumbing the whole app. In general, since these changes need to be maintainable long-term, we prefer smaller, cumulative PRs that can be individually reviewed and tested to large complex changes that may have unexpected consequences.

rksh:
condensed to https://github.com/wiglenet/wigle-wifi-wardriving/pull/834 for the bugfix. Next steps might include display improvements, data storage enhancement.
For changes to the network model and CSV format, https://api.wigle.net/csvFormat-1_6.html is the current published spec

PR #833, opened 2026-05-02
fix: reliability fix for play console errors proposed by the basilisk

rksh:
AI-generated deep cut to get rid of pre-androidx.fragment 1.8.3 crash potential
https://developer.android.com/jetpack/androidx/releases/fragment#1.8.3


bobzilladev (approved):
Spooky

PR #835, opened 2026-05-04
fix: bad FOSS map url fix

rksh:
Seeing crashes on FOSS map URLs in Play Console. Should be propagated to foss-main as well.

### DIFF
--- a/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/SettingsFragment.java
+++ b/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/SettingsFragment.java
@@ -805,53 +805,68 @@ public final class SettingsFragment extends Fragment implements DialogListener {
             verify.setOnClickListener(buttonView -> {
                 final Editable urlPart = fossMapStyleUrlEdit.getText();
                 final Editable keyPart = fossMapKeyEdit.getText();
-                if (null == urlPart || urlPart.toString().isEmpty()) {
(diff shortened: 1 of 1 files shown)


wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/SettingsFragment.java
```diff
@@ -804,51 +805,68 @@ private void setupFossMapEditFields(final View view, final SharedPreferences pre
...
-                            mainHandler.post(() -> {
-                                fossMapStyleUrlEdit.setError("required");
+                    final String concat = urlCandidate + keyCandidate;
+                    final HttpUrl httpUrl = HttpUrl.parse(concat);
+                    if (httpUrl == null) {
+                        fossMapStyleUrlEdit.setError("invalid");
+                    } else {
+                        final String scheme = httpUrl.scheme();
+                        if (!"http".equals(scheme) && !"https".equals(scheme)) {
+                            fossMapStyleUrlEdit.setError("invalid");
```
bobzilladev: could say invalid scheme and include the scheme, if yer feeling nice

bobzilladev (approved):
one nit

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/SettingsFragment.java
rksh: what an excellent point. added!

Issue #837, opened 2026-05-17
[Bug] Address search performs synchronous `Geocoder#getFromLocationName()` on the UI thread


venkyqz:
Hi Wiglnet Team,

I’m a PhD student researching Android thread-related issues. My research group recently ran a static-analysis scan for thread-related bugs in real-world F-Droid apps, and our prototype flagged a potential issue in WiGLE WiFi Wardriving.

## Checked target

- **Source-level caller:** `net.wigle.wigleandroid.util.SearchUtil.setupQuery(View, Context, boolean)`
- **Detected API / pattern:** synchronous `Geocoder#getFromLocationName(String, int)`
- **Observed context:** main/UI thread, through the search button click path
- **Expected context:** use the asynchronous Geocoder API where available, or run synchronous geocoding on a background thread

## What I found

The search UI registers a click listener for the search button. Inside that listener, the code calls:

```java
String fail = SearchUtil.setupQuery(view, getActivity(), local);
```

`SearchUtil.setupQuery(...)` then processes the address query field. When the current field is `R.id.query_address`, it constructs a `Geocoder` and performs a synchronous address lookup:

```java
Geocoder gc = new Geocoder(context);
List<Address> addresses = gc.getFromLocationName(text, 1);
```

This call is reached directly from the UI click listener path, so a user-submitted address search can block the main thread while geocoding runs.

I also noticed similar synchronous geocoding patterns in the map/address search code paths, such as `FossSearchFragment.setupAddressSearch(...)` and `SearchFragment.setupAddressSearch(...)`. Those may be worth fixing in the same patch, although the static-analysis finding above specifically points to `SearchUtil.setupQuery(...)`.

## Verified bug trace

```text
User taps the search button
  -> View.OnClickListener runs on the main/UI thread
  -> AbstractSearchFragment.setupQueryButtons(...)
  -> SearchUtil.setupQuery(view, getActivity(), local)
  -> address field branch: R.id.query_address
  -> new Geocoder(context)
  -> geocoder.getFromLocationName(text, 1)
  -> synchronous geocoding is performed on the UI thread
```

## Why this matters

Android documents the synchronous `Geocoder#getFromLocationName(...)` overload as potentially blocking. The API may perform network access and may block for an excessive amount of time. The documentation strongly encourages using the asynchronous geocoder API to avoid blocking a thread, and says that if the synchronous API must be used, it should be run on a background thread.

[...]

PR #838, opened 2026-05-17
fix: for potential ANR in Android 33+ - #837

rksh:
- breaks Geocoding calls into a utility class
- defines a callback interface
- uses that callback interface in classes w/ geocoding calls


bobzilladev (approved):
Nice

Issue #846, opened 2026-06-06
WigleWifi error report


Tom60chat:
WigleWifi error log - 6 juin 2026 12:38:39
versionName: 2.108
packageName: net.wigle.wigleandroid
MODEL: SM-S938W
RELEASE: 16
BOARD: sun
BRAND: samsung
DEVICE: pa3q
DISPLAY: BP2A.250605.031.A3.S938WVLS9BZBH
FINGERPRINT: samsung/pa3qcsx/pa3q:16/BP2A.250605.031.A3/S938WVLS9BZBH_OYV9BZBH:user/release-keys
HOST: 21DND515
ID: BP2A.250605.031.A3
PRODUCT: pa3qcsx
TAGS: release-keys
TIME: 1774941735000
TYPE: user
USER: dpi
baseError: MapLibreConfigurationException: 
Using MapView requires calling MapLibre.getInstance(Context context, String apiKey, WellKnownTileServer wellKnownTileServer) before inflating or creating the view.

java.lang.RuntimeException: Unable to start activity ComponentInfo{net.wigle.wigleandroid/net.wigle.wigleandroid.FossDBResultActivity}: android.view.InflateException: Binary XML file line #18 in net.wigle.wigleandroid:layout/foss_dbresult: Binary XML file line #18 in net.wigle.wigleandroid:layout/foss_dbresult: Error inflating class org.maplibre.android.maps.MapView
	at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:4710)
	at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:4941)
	at android.app.servertransaction.LaunchActivityItem.execute(LaunchActivityItem.java:222)
	at android.app.servertransaction.TransactionExecutor.executeNonLifecycleItem(TransactionExecutor.java:133)
	at android.app.servertransaction.TransactionExecutor.executeTransactionItems(TransactionExecutor.java:103)
	at android.app.servertransaction.TransactionExecutor.execute(TransactionExecutor.java:80)
	at android.app.ActivityThread$H.handleMessage(ActivityThread.java:3150)
	at android.os.Handler.dispatchMessage(Handler.java:110)
	at android.os.Looper.loopOnce(Looper.java:273)
	at android.os.Looper.loop(Looper.java:363)
	at android.app.ActivityThread.main(ActivityThread.java:10060)
	at java.lang.reflect.Method.invoke(Native Method)
[...]

PR #847, opened 2026-06-07
fix: Inflate exception foss map config

rksh:
This shouldn't happen with the validation introduced in 109, but belt-and-suspenders. 

Addresses #846


bobzilladev (approved):
Tip toeing through a mine field

PR #848, opened 2026-06-19
fix: speculative NPE fix; could this even happen?

rksh:
getting an NPE on this line in the play console - seems highly unlikely, but see anything else that could cause it?


bobzilladev (approved):
This is what Brian Boitano would do.

PR #852, opened 2026-07-10
fix: trying to balance excessive wakeups warning

rksh:
co-developed with the basilisk.


bobzilladev (approved):
Willing to try any dark magicks to make it just do the work the user is asking for!

PR #856, opened 2026-07-26
[FIX] Ongoing memory use improvements

rksh:
Google maps memory management clean-up (parallel w/ FOSS)

going to do more extensive play-testing with this before merging.


bobzilladev (approved):
so much lifecycle rigmarole, thank you for tackling

PR #857, opened 2026-07-31
feat: optional per-row RSSI background histogram visualization in list view

rksh:
little dynamic graphs of signal strength over time!
off by default, since this has perf. consequences


bobzilladev (approved):
approved as part of: https://github.com/wiglenet/wigle-wifi-wardriving/pull/858

PR #862, opened 2026-08-04
fix: choose dashboard layout by window width, not device orientation

rksh:
Orientation stopped doesn't do what it used to do anyway. This is the new way of doing it, I guess?

Also inflate the halves lazily rather than building both on every onCreateView, which requires clearing them in onDestroyView so a cached half isn't re-added while it still holds the outgoing ScrollView as parent.


bobzilladev (approved):
How responsive of you

PR #863, opened 2026-08-04
Alert export

rksh:
no import yet, just export - but it makes it easier to share filters.

### DIFF
--- a/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/model/Filter.java
+++ b/wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/model/Filter.java
@@ -11,14 +11,14 @@ import java.util.List;
 import java.util.Map;
 
 /**
- * Export-/Import-able network matching filter.
+ * Export-/Import-able network matching filter, carried in a {@link FilterSet}.
  * Provides description and a list of criteria to use in matching / exclusion from matching.
  * Serialize/deserialize via {@link net.wigle.wigleandroid.util.FilterJson}.
  *
  * <p>Sample JSON payload:
  * <pre>{@code
+ *
  * {
- *   "schemaVersion": 1,
  *   "description": "exclude AirTags",
  *   "exclude": true,
  *   "macAddresses": ["AA:BB:CC:DD:EE:FF"],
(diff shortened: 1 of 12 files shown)


wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/model/Filter.java
```diff
@@ -0,0 +1,556 @@
...
+import java.util.Collection;
+import java.util.Collections;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Export-/Import-able network matching filter.
+ * Provides description and a list of criteria to use in matching / exclusion from matching.
+ * Serialize/deserialize via {@link net.wigle.wigleandroid.util.FilterJson}.
```
bobzilladev, line 16: I think this makes sense for a given rule, but also that there should be a package to have multiple rules together as a full filter set. Just some kind of notion of input/output n things.

bobzilladev (approved):
One comment

wiglewifiwardriving/src/main/java/net/wigle/wigleandroid/model/Filter.java
rksh: so, I'm juggling the difference between "use this filter for alerts" vs. (eventually) "use this filter for exclusion from logging" &etc. I don't know the whole space of possible uses yet, so I was keeping this modular. That being said, it can have as many MACs, OUIs, BLE MFGR IDs, and BLE UUID any/all sets as desired. Say more?

bobzilladev: This just all feels like an access control list, and so I'd lean into it. You want a large set with a more general wildcard, then add some excludes for sections of the space you don't control or don't care about, you can turn on and off individual rules if you want to test with or without something specific, you can sync an entire ruleset and users can decide to toggle stuff on and off that they care about, that kind of vibe.

rksh: since this is currently just for sounding alarms (and has to implemented reliably on a per-scan), I was going for simple matching rules. this might require a design session

bobzilladev (approved):

PR #864, opened 2026-08-06
feat: Network activity rssi graph overlay

rksh:
adds dynamic RSSI graphs to bottom chrome in network activity IFF you have RSSI graphing enabled
also, does some layout juggling for landscape mode because things were getting dumb


bobzilladev (approved):
⚡

PR #865, opened 2026-08-15
DB contention fixes and modernization

rksh:
Submitting as a **_draft_** for now - this is going to need a ton of testing.
Short version: database contention is a common cause of play console crashes - revisiting old mode and locking assumptions from the early days and trying to modernize them.
- switching to WAL mode to make logging more reliable/efficient
- handling a bunch of cases with timeouts and contention explicitly
- avoiding multiple helpers/queues that could result in deadlocks/contention


rksh:
putting this out as ready for review - tested/working on multiple devices over the last month, more testing certainly required.

bobzilladev (approved):
Have been running for a few weeks, good afaict!

